PR #26, opened 2024-09-17
Rename base-account to service-account

stiiifff:


stiiifff:
Discussed this with @Art3miX. Closing as this is probably not a good idea.

PR #31, opened 2024-09-18
Forwarder service

stiiifff:
Opening the PR for an early review.

Implemented:

- Implementation (mostly) according to [spec](https://www.notion.so/Forwarder-service-specification-for-review-680677e9d5cb4cebad36d23987107eb2).
- Supports native & CW20 tokens via the use of the `cw-denom` crate.
- Tests, including the scenario of continuously forwarding funds

Todo:
- [x] Update the README ..
- [x] Add support for updating the configuration (+ tests).
- [x] Extract some re-usable types & test logic into `service-base` and `service-utils`.
- [x] Make the test code cleaner
- [x] ~~Reply to processor when an `execution_id` has been specified.~~

Notes: there is a bit more code than anticipated in the `msg.rs` due to the unchecked / checked nature of denoms, and validation logic to ensure that the specified denoms are unique.


contracts/services/forwarder/src/contract.rs
```diff
@@ -0,0 +1,153 @@
...
+                    .map(|(amount, denom)| denom.get_transfer_to_message(cfg.output_addr(), amount))
+                    .collect::<StdResult<Vec<CosmosMsg>>>()?;
+
+                // Wrap the transfer messages to be executed on behalf of the input account
+                let input_account_msgs = execute_on_behalf_of(transfer_messages, cfg.input_addr())?;
+
+                // Save last successful forward
+                LAST_SUCCESSFUL_FORWARD.save(deps.storage, &env.block)?;
+
+                if let Some(_execution_id) = execution_id {
```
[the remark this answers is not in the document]
stiiifff: I removed it

contracts/services/forwarder/src/msg.rs
```diff
@@ -0,0 +1,247 @@
...
+    Forward {},
+}
+
+#[cw_serde]
+#[derive(QueryResponses)]
+/// Enum representing the different query messages that can be sent.
+pub enum QueryMsg {
+    /// Query to get the owner address.
+    #[returns(Addr)]
+    GetOwner {},
```
keyleu, line 27: the cw_ownable crate has a #[cw_ownable_query] macro that will add the ownership query here. I'd suggesting using that instead of redefining it here.

contracts/services/forwarder/src/msg.rs
```diff
@@ -0,0 +1,247 @@
...
+
+impl ServiceConfigValidation<Config> for ServiceConfig {
+    fn validate(&self, deps: Deps) -> Result<Config, ServiceError> {
+        let input_addr = deps.api.addr_validate(&self.input_addr)?;
+        let output_addr = deps.api.addr_validate(&self.output_addr)?;
+        // Ensure denoms are unique in forwarding configs
+        ensure_denom_uniqueness(&self.forwarding_configs)?;
+
+        // Convert the unchecked denoms to checked denoms
+        let checked_fwd_configs = convert_to_checked_configs(&self.forwarding_configs, deps)?;
```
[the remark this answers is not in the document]
stiiifff: Yeah, I thought about `TryInto` but the conversion logic needs the `deps`.

contracts/services/forwarder/src/msg.rs
```diff
@@ -0,0 +1,247 @@
...
+        }
+    }
+}
+
+impl ServiceConfigValidation<Config> for ServiceConfig {
+    fn validate(&self, deps: Deps) -> Result<Config, ServiceError> {
+        let input_addr = deps.api.addr_validate(&self.input_addr)?;
+        let output_addr = deps.api.addr_validate(&self.output_addr)?;
+        // Ensure denoms are unique in forwarding configs
+        ensure_denom_uniqueness(&self.forwarding_configs)?;
```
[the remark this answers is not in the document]
stiiifff: Could be .. do you feel that it's needed ? or a simple func does the job .. ?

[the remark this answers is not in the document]
keyleu: both do the job, was just an idea

keyleu (approved):

packages/service-utils/src/lib.rs
```diff
@@ -50,11 +55,13 @@ impl ServiceAccountType {
     }
 }
 
+// Private enum mimicking the ExecuteMsg from the base_account contract
 #[cw_serde]
-pub enum ExecuteMsg {
+enum ExecuteMsg {
     ExecuteMsg { msgs: Vec<CosmosMsg> }, // Execute any CosmosMsg (approved services or admin)
 }
```
Art3miX, line 62: I don't think mocking is good here, very error prone, if base account changes its api, will be hard to debug.
Maybe its good for now, not sure how accounts will be played in the future, but will be safer to use the execute enum directly from the account contract here.

contracts/services/forwarder/src/tests.rs
```diff
@@ -98,192 +48,105 @@ impl ForwarderTestSuite {
...
-        self.app.api()
-    }
-
-    fn instantiate(&mut self, svc_cfg: &ServiceConfig) -> AnyResult<Addr> {
-        let msg = InstantiateMsg {
-            owner: self.owner.to_string(),
-            processor: self.processor.to_string(),
-            config: svc_cfg,
+    pub fn forwarder_init(&mut self, cfg: &ServiceConfig) -> Addr {
+        let init_msg = InstantiateMsg {
```
Art3miX, line 63: My personal preference is to use the whole type path for contract common types, make its much easier to understand of which contract this type is, without "verifying" this is the correct type.
Less necessary here, because of the function name, but still a bit nicer.

contracts/services/forwarder/src/tests.rs
```diff
@@ -98,192 +48,105 @@ impl ForwarderTestSuite {
...
+            assert_eq!(account_addr, input_addr);
+
+            if let Some(balances) = self.input_balances.as_ref().cloned() {
+                let amounts = balances
+                    .iter()
+                    .map(|(amount, denom)| coin(*amount, denom.to_string()))
+                    .collect::<Vec<Coin>>();
+                self.init_balance(&input_addr, amounts);
+            }
+        }
```
Art3miX, line 82: Can you explain a little about this?

contracts/services/forwarder/src/tests.rs
```diff
@@ -98,192 +48,105 @@ impl ForwarderTestSuite {
...
+    pub fn forwarder_init(&mut self, cfg: &ServiceConfig) -> Addr {
+        let init_msg = InstantiateMsg {
+            owner: self.owner().to_string(),
+            processor: self.processor().to_string(),
+            config: cfg.clone(),
         };
+        let addr = self.contract_init(self.forwarder_code_id, "forwarder", &init_msg, &[]);
+
+        let input_addr = self.input_addr().clone();
+        if self.inner.app_mut().contract_data(&input_addr).is_err() {
```
Art3miX, line 71: We have `self.app_mut`

Art3miX (approved):
just few comments, nothing major

contracts/services/forwarder/src/tests.rs
stiiifff: I see what you mean, I agree. But probably less of a big deal here as the tests are in the contract's crate.

packages/service-utils/src/lib.rs
stiiifff: Yeah, I am not really satisfied with this solution either actually.

contracts/services/forwarder/Cargo.toml
```diff
@@ -33,3 +34,4 @@ cw20                 = { workspace = true }
 cw20-base            = { workspace = true }
 sha2                 = { workspace = true }
 valence-base-account = { workspace = true }
+valence-service-utils = { workspace = true, features = ["testing"] }
```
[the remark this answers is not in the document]
stiiifff: fixed

packages/service-utils/src/lib.rs
stiiifff: I moved the `valence-base-account` 's `msg` & `error` to a new `valence-account-utils` package.

contracts/services/forwarder/src/tests.rs
stiiifff: Create the input account if it hasn't been created yet, and initialise its balance. The initial check with `contract_data` ensures we only do this once even if the function is mistakenly called multiple times.

stiiifff:
LGTM !

PR #36, opened 2024-09-19
local-ic CI

keyleu:
Adds local-ic to CI. After making our repo public, CI doesn't fail during contract uploading anymore, which was the most common error before, which confirms the hardware issues. Got 1 error during wasm connection creation but I was spamming with tests and got rate-limited so that's probably why.

For now I've stabilized it quite a lot playing with the time out values and restarting the relayer in the code when it doesn't process packets but even sometimes that's not enough.

Discussed with @uditvira to include it into CI even if we get the occasional error due to rate-limiting. If we see we are getting too many errors, we can just remove it again, and maybe wait until we get a dedicated runner to have closer to 100% success rate.


Art3miX (approved):

keyleu:
Will leave this one open for a while to verify it stabilized

PR #37, opened 2024-09-20
Process tokenfactory tokens during processor callback

### LINKED ISSUE #19: Handle tokenfactory burn/return when getting a processor callback
Right now we aren't dealing with the tokenfactory token when we receive a processor callback. We should correctly burn or return it to the user depending on the Execution result and update the current executions

keyleu:
Closes #19 .

If a user who spent a token to execute a batch (permissioned with limit) couldn't execute anything, his token is given back instead of being consumed.

### DIFF
--- a/contracts/authorization/src/contract.rs
+++ b/contracts/authorization/src/contract.rs
@@ -712,7 +712,14 @@ fn process_processor_callback(
     CURRENT_EXECUTIONS.update(
         deps.storage,
         callback.label.clone(),
-        |current| -> StdResult<_> { Ok(current.unwrap_or_default().saturating_sub(1)) },
+        |current| -> Result<u64, ContractError> {
+            let count = current.unwrap_or_default();
+            if count == 0 {
+                Err(ContractError::CurrentExecutionsIsZero {})
+            } else {
+                Ok(count - 1)
+            }
+        },
     )?;
 
     // Check if a token was sent to perform this operation and that it wasn't started by the owner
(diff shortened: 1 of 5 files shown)


contracts/authorization/src/contract.rs
```diff
@@ -693,12 +704,47 @@ fn process_processor_callback(
...
+    // Update the result
     callback.execution_result = execution_result;
-
     PROCESSOR_CALLBACKS.save(deps.storage, execution_id, &callback)?;
 
+    // Reduce the current executions for the label
+    CURRENT_EXECUTIONS.update(
+        deps.storage,
+        callback.label.clone(),
+        |current| -> StdResult<_> { Ok(current.unwrap_or_default().saturating_sub(1)) },
```
Art3miX, line 715: This looks weird, default for u64 is 0, seems like if its 0 it should error, no?
Can you explain a little bit more about it?

contracts/authorization/src/contract.rs
```diff
@@ -693,12 +704,47 @@ fn process_processor_callback(
...
+    )?;
+
+    // Check if a token was sent to perform this operation and that it wasn't started by the owner
+    let authorization = AUTHORIZATIONS.load(deps.storage, callback.label.clone())?;
+    let mut messages = vec![];
+    if let OperationInitiator::User(initiator_addr) = &callback.initiator {
+        if let AuthorizationMode::Permissioned(PermissionType::WithCallLimit(_)) =
```
Art3miX, line 722: Can you explain more about user vs owner here? 
Owners basically are free to do any operation for free?

contracts/authorization/src/contract.rs
keyleu: It won't be 0, if we are processing a callback it will be at least 1 or more. This tracks how many are being executed for each label.

contracts/authorization/src/contract.rs
keyleu: Yes, Owner can insert / remove actions from the any queue for free without restrictions

contracts/authorization/src/contract.rs
Art3miX: If it should never be 0, we should test for it here.

keyleu: Sure, added it

Art3miX (approved):

PR #38, opened 2024-09-23
Add polytone tokenfactory handling and correct ExecutionResult update

### LINKED ISSUE #23: Refund tokens when polytone timeouts and ttl is expired
We should refund authorization tokens (those that were sent to the contract to trigger messages) if there is a timeout and ttl is expired or when we try to retry timed out messages and the ttl is expired

keyleu:
Closes #23 

Final update to do correct handling of Polytone callbacks and retrying.

Now, every time we receive a callback, we check if it's going to be retriable or not. If it isn't, we immediately set the retriable flag to false and return the token if there is a token to return.
During retries, now we check if it's currently retriable and that TTL hasn't expired. If it's marked as retriable but the TTL has expired, we send the token back and set the flag to false to not be able to retry again (This applies to the case that when we receive the callback it's still retriable, but when we are about to retry it's not retriable anymore).

It also fixes an issue where we were not correctly updating the status to `InProcess` when we were retrying. Added a particular test to catch this as well.

Additionally added all the tests to local-ic to verify the balances and to correctly see that TTL handling is done correctly.


contracts/authorization/src/contract.rs
```diff
@@ -614,30 +610,50 @@ fn retry_msgs(
...
-            create_wasm_msg_for_processor_or_bridge(
+            messages.push(create_msg_for_processor_or_bridge(
                 deps.storage,
                 execute_msg_binary,
                 &callback_info.domain,
                 Some(callback_request),
-            )?
+            )?);
         }
         _ => {
```
Art3miX, line 628: why `_`? it can only be None, being ambiguous here has to benefit and can actually be worse

contracts/authorization/src/contract.rs
```diff
@@ -803,11 +821,40 @@ fn process_polytone_callback(
...
+                                    let denom = build_tokenfactory_denom(
+                                        env.contract.address.as_str(),
+                                        &callback_info.label,
+                                    );
+                                    messages.push(CosmosMsg::Bank(BankMsg::Send {
+                                        to_address: initiator_addr.to_string(),
+                                        amount: coins(1, denom),
+                                    }));
+                                }
+
```
Art3miX, line 857: This is written in a somewhat confusing way, the main reason is that you check if `callback_info.execution_result` is in process, change that to something else, and then check again if its ExecutionResult::Timeout, all in the same block of code.
Would be much more clear for future readers if you move the 2nd `if let` code into the first one: "if error is timeout and is expired, we refund the token and set result to be timeout expired", a much more natural flow.

keyleu: Agree, changed it to what you specified to make it more natural.

contracts/authorization/src/contract.rs
keyleu: It can not only be None, that's why I used _.
First case is `Some(ttl) if !ttl.is_expired(&env.block)`
It can be None or Some and be expired. For both possible cases I want the same action

Art3miX: Right, forgot this match syntax

Art3miX (approved):

PR #42, opened 2024-09-26
Update Base account contract

### LINKED ISSUE #18: Change base account to cw-ownable
Currently its using our own implementation of ownership, for consistency we would like to change it to cw-ownable

stiiifff:
Use cw-ownable instead of custom implementation.
Created an `account-utils` package with shared logic testing for account contracts, including some test primitives.
Added tests for: approving / removing services from the account, transferring tokens (native or cw20) from account's balance by the owner or an approved service, transferring or renouncing ownership.

Note: there's now redundant code between the `account-utils` and `service-utils` `testing` module. I suggest moving this to a `testing-utils` package in a separate PR.

closes #18


keyleu (approved):

keyleu:
Looks good! Not sure if all the cw_ownable tests were needed because it's a widely used crate but good to have anyways

PR #44, opened 2024-09-26
Astroport Withdrawer combined

keyleu:
Much simpler than LPer, combines the liquidity withdrawer in one contract.


contracts/services/astroport-withdrawer/src/contract.rs
```diff
@@ -0,0 +1,150 @@
...
+                ));
+            }
+            let config = CONFIG.load(deps.storage)?;
+            actions::process_action(deps, env, info, action_msgs, config)
+        }
+        ExecuteMsg::UpdateConfig { new_config } => {
+            cw_ownable::assert_owner(deps.as_ref().storage, &info.sender)?;
+            let config = new_config.validate(deps.as_ref())?;
+            CONFIG.save(deps.storage, &config)?;
+            Ok(Response::new().add_attribute("method", "update_config"))
```
Art3miX, line 57: Same config thing

Art3miX (changes requested):
Some comments from previous PRs apply here as well, didn't want to duplicate them.

contracts/services/astroport-withdrawer/src/contract.rs
keyleu: Refactored contract

contracts/services/astroport-withdrawer/src/astroport_cw20.rs
```diff
@@ -0,0 +1,72 @@
...
+        cfg.pool_addr.clone(),
+        &valence_astroport_utils::astroport_cw20_lp_token::PoolQueryMsg::Share {
+            amount: balance_response.balance,
+        },
+    )?;
+
+    // Create the withdraw and send messages
+    let withdraw_msg = CosmosMsg::Wasm(cosmwasm_std::WasmMsg::Execute {
+        contract_addr: token_addr.to_string(),
+        msg: to_json_binary(&Cw20ExecuteMsg::Send {
```
stiiifff, line 48: I don't understand this part .. can you explain how it works ?

keyleu: Yes, in Cw20 we have both Transfer and Send.
- Transfer is similar to a bank send message, just send the CW20 token to an address.
- Send on the other hand, is to send CW20 tokens to a contract and atomically trigger a hook on that contract (in this case, the withdrawliquidity message that astroport defines in its pool contract that I specify below)

In summary, what I'm saying is: Send these tokens (they are cw20, so it's a smart contract execution) to this contract address and immediately trigger the action (hook) that I'm specifying here. Astroport will receive those cw20 tokens and do the action in the contract.

If it's still not clear lmk and we can hop on a quick call.

stiiifff: Understood, all clear, thanks !

stiiifff (approved):

contracts/services/astroport-withdrawer/README.md
```diff
@@ -0,0 +1,57 @@
...
+## Configuration
+
+The service is configured on instantiation via the `ServiceConfig` type.
+
+```rust
+pub struct ServiceConfig {
+    // Account from which the funds are LPed
+    pub input_addr: String,
+    // Account to which the LP tokens are forwarded
+    pub output_addr: String,
```
Art3miX, line 40: Use `ServiceAccountType` here

Art3miX (approved):

PR #47, opened 2024-09-30
Finish workflow init

Art3miX:
Finish with all the code that is required for the workflow init process.
Still missing tests, so this is probably not 100% working right now, need to test and make sure the communication with the chain is working (manually and write automatic tests)


Art3miX:
I've added sleep after each execute or instantiate call to the chain, because we have multiple of them, and we have no way of getting feedback when this message was actually executed, we need to rely on sleep here, which is a really bad design.

We can have separate queue of messages need to be executed, and execute them in batches, but we would still need to apply sleep between batches, and would need to handle retries for specific messages.

workflow-manager/src/domain/cosmos_cw.rs
```diff
@@ -325,18 +329,23 @@ impl Connector for CosmosCosmwasmConnector {
...
-            // .broadcast_tx(vec![m], None, None, BroadcastMode::Sync) // TODO: change once we ready
+            // .simulate_tx(vec![m])
+            .broadcast_tx(vec![m], None, None, BroadcastMode::Sync)
             .await
-            .map(|_| ())
-            .context("Failed to broadcast the TX")
-            .map_err(|e| CosmosCosmwasmError::Error(e).into())
+            .map_err(CosmosCosmwasmError::Error)?;
+
+        sleep(std::time::Duration::from_secs(12)).await;
```
keyleu, line 337: question regarding this, can't we query the TX hash?
this is how local-IC does it for example
```rust
        for _ in 0..5 {
            let data = self.query_tx_hash(&tx_hash);

            if !data.to_string().starts_with("{\"error\":") {
                return Ok(data);
            }

            std::thread::sleep(std::time::Duration::from_secs(1));
        }

File: workflow-manager/src/domain/cosmos_cw.rs [hunk 2]
```diff
@@ -652,6 +676,148 @@ impl Connector for CosmosCosmwasmConnector {
...
+        &mut self,
+        main_domain: &str,
+        id: u64,
+    ) -> ConnectorResult<valence_workflow_registry_utils::WorkflowResponse> {
+        if self.chain_name != main_domain {
+            return Err(CosmosCosmwasmError::Error(anyhow::anyhow!(
+                "Adding external domain is only possible on main domain in authorization contract"
+            ))
+            .into());
+        }
```
keyleu, line 690: question: workflow registry will always be in Main domain right?

regarding error, shouldn't the error be different in this case as we are not adding external domain here (maybe copy pasted from a different place)

workflow-manager/src/domain/cosmos_cw.rs
```diff
@@ -848,4 +1015,44 @@ impl CosmosCosmwasmConnector {
...
+            .map_err(CosmosCosmwasmError::Error)?
+            .code_id)
+    }
+
+    pub async fn get_contract_name_by_code_id(
+        &mut self,
+        addr: String,
+    ) -> Result<String, CosmosCosmwasmError> {
+        let code_id = self.get_code_id_of_addr(addr.clone()).await?;
+
```
keyleu, line 1047: nit: was a bit confused here. Name is get_contract_name_by_code_id but you are actually passing an address as argument and getting the code_id inside

workflow-manager/src/domain/cosmos_cw.rs
```diff
@@ -652,6 +676,148 @@ impl Connector for CosmosCosmwasmConnector {
...
+            .context("'verify_account' Code id doesn't match any account type")
+            .map(|_| ())
+            .map_err(CosmosCosmwasmError::Error)?)
+    }
+
+    async fn verify_processor(&mut self, processor_addr: String) -> ConnectorResult<()> {
+        let contract_name = self.get_contract_name_by_code_id(processor_addr).await?;
+
+        // Make sure the code id is of name processor
+        if contract_name != "processor" {
```
keyleu, line 749: nit: should we use valence-processor, to stay consistent on contracts/workflow manager?

workflow-manager/src/domain/cosmos_cw.rs
```diff
@@ -652,6 +676,148 @@ impl Connector for CosmosCosmwasmConnector {
...
+                "'verify_processor' Code id isn't of processor code id"
+            ))
+            .into());
+        }
+
+        Ok(())
+    }
+
+    async fn verify_bridge_account(&mut self, bridge_addr: String) -> ConnectorResult<()> {
+        // If the address have a code id, it means it was instantiated.
```
keyleu, line 760: typo *has

workflow-manager/src/domain/cosmos_cw.rs
```diff
@@ -652,6 +676,148 @@ impl Connector for CosmosCosmwasmConnector {
...
+        self.get_code_id_of_addr(bridge_addr)
+            .await
+            .map(|_| ())
+            .map_err(|e| e.into())
+    }
+
+    async fn verify_authorization_addr(&mut self, addr: String) -> ConnectorResult<()> {
+        let code_id = *self
+            .code_ids
+            .get("authorization")
```
keyleu, line 770: nit: same as above, use valence-authorization?

workflow-manager/src/workflow_config.rs
```diff
@@ -411,9 +433,78 @@ impl WorkflowConfig {
...
+    ) -> ManagerResult<()> {
+        let mut neutron_connector = connectors.get_or_create_connector(&NEUTRON_DOMAIN).await?;
+        // verify id that is used in workflow is not taken and is not 0
+        ensure!(
+            neutron_connector
+                .query_workflow_registry(NEUTRON_DOMAIN.get_chain_name(), self.id)
+                .await
+                .is_ok(),
+            ManagerError::WorkflowIdAlreadyExists(self.id)
+        );
```
keyleu, line 451: We are not checking if it's not 0, right? Change comment/check

workflow-manager/src/domain/cosmos_cw.rs
Art3miX: Yes we can, I wanted to make sure i unblock the workflow, and const waiting time is the fastest for now.

Art3miX: Opened an issue to not forget #50

workflow-manager/src/domain/cosmos_cw.rs
Art3miX: Yup, started with a single function and separated them.
Fixed.

workflow-manager/src/domain/cosmos_cw.rs
Art3miX: I would like not to, but I can see why that might be preferred, we already remove valence prefix in our folder names for example.
The reason I prefer not to, is because this naming is not just for the config, rather its also used in enums for example, keep it as simple as possible is best in this case.

@stiiifff @bekauz Thoughts?

workflow-manager/src/workflow_config.rs
Art3miX: Fixed

Art3miX:
@keyleu Made the changes in the new PR, part of the test changes.

keyleu (approved):

workflow-manager/src/domain/cosmos_cw.rs
stiiifff: This seems a bit weak in terms of verification .. maybe query the contract's cw2 info (and verify the contract's crate name) would be better ?

workflow-manager/src/domain/cosmos_cw.rs
stiiifff: @keyleu @Art3miX Yeah the sleep is not great, nor the looping + sleep tbh. I think the correct pattern would be to use a wait handle to block the current thread until either, the transaction is finalized or errors out (need to subscribe to chain events before broadcasting the tx .. I'm not sure what's the correct API for that), or there is a timeout of x seconds. Basically, you want to wait asynchronously until some events occur, not wait synchronously and poll for result.

PR #49, opened 2024-10-03
Reverse Splitter service

stiiifff:
Implemented:

- Implementation according to [spec](https://www.notion.so/Reverse-Splitter-service-specification-for-review-cb2a7e0283a54b6fb47920aaf8734198)
- Support for fixed amounts or ratios
- Supports native & CW20 tokens
- Extensive tests

Todo:
- [x] Dynamic ratio support via an external contract
- [x] Update ServiceConfig to use ServiceAccountType for input & output accounts


contracts/services/reverse-splitter/src/msg.rs
```diff
@@ -0,0 +1,432 @@
...
+                            .to_string(),
+                    ));
+                }
+            }
+            (None, Some(UncheckedRatioConfig::FixedRatio(ratio))) => {
+                if split.denom == *base_denom && *ratio != Decimal::one() {
+                    return Err(ServiceError::ConfigurationError(
+                        "Invalid split config: fixed ratio for base denom must be 1.".to_string(),
+                    ));
+                }
```
keyleu, line 372: Just thought about this and probably applies to splitter too. Is it worth validating ratios so that they are 0 < ratio <= 1?

contracts/services/reverse-splitter/src/contract.rs
```diff
@@ -0,0 +1,299 @@
...
+                            let factor = split.factor();
+                            if let Some(factor) = factor {
+                                amount = amount
+                                    .checked_div((*factor as u128).into())
+                                    .map_err(|err| ServiceError::Std(err.into()))?;
+                            }
+                            Ok((amount, *ratio, factor, denom, account))
+                        })
+                    })
+                    .expect("Split config must have either an amount or a ratio")
```
keyleu, line 209: Nit: This is unreachable because we validated it right? Might add it here in a comment or something

contracts/services/reverse-splitter/src/contract.rs
```diff
@@ -0,0 +1,299 @@
...
+            .iter()
+            .map(|split| {
+                // Lookup account/denom balance
+                let account = split.account();
+                let denom = split.denom();
+                let balance = account_balances.get(&account_key(account, denom)).unwrap();
+                split
+                    .amount()
+                    .map(|amount| {
+                        assert_eq!(denom, cfg.base_denom());
```
keyleu, line 184: remove this? It has been already validated no? Or if you want to double check return a ServiceError instead?

contracts/services/reverse-splitter/src/contract.rs
stiiifff: Yes, it is an invariant .. could have done an unwrap() instead.

contracts/services/reverse-splitter/src/contract.rs
stiiifff: Yep .. I'll remove it.

contracts/services/reverse-splitter/README.md
```diff
@@ -0,0 +1 @@
+# Valence Reverse Splitter service
```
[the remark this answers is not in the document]
stiiifff: For some reason, I thought I had added the content to the readme but it's not case .. probably lost it in a stash 😅

contracts/services/reverse-splitter/src/msg.rs
stiiifff: I've added a test to check for non-zero amount & ratio. For the reverse splitter, the ratio doesn't have to be between 0 and 1 (it can be an exchange rate for the base denom). For the Splitter, the check on the sum of ratios being 1 is sufficient.

keyleu (approved):

Cargo.toml
```diff
@@ -49,11 +49,11 @@ sha2            = "0.10.8"
...
-valence-processor          = { path = "contracts/processor", features = ["library"] }
-valence-reverse-splitter   = { path = "contracts/services/reverse-splitter-temp", features = ["library"] }
-valence-splitter-service   = { path = "contracts/services/splitter", features = ["library"] }
+valence-authorization            = { path = "contracts/authorization", features = ["library"] }
+valence-base-account             = { path = "contracts/accounts/base_account", features = ["library"] }
+valence-processor                = { path = "contracts/processor", features = ["library"] }
+valence-reverse-splitter-service = { path = "contracts/services/reverse-splitter", features = ["library"] }
+valence-splitter-service         = { path = "contracts/services/splitter", features = ["library"] }
 valence-test-dynamic-ratio = { path = "contracts/testing/test-dynamic-ratio", features = ["library"] }
 valence-test-service       = { path = "contracts/testing/test-service", features = ["library"] }
```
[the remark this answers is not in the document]
stiiifff: fixed

contracts/services/reverse-splitter/Cargo.toml
```diff
@@ -0,0 +1,37 @@
+[package]
+name    = "valence-reverse-splitter-service"
+authors    = { workspace = true }
+edition    = { workspace = true }
```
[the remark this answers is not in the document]
stiiifff: fixed

contracts/services/reverse-splitter/src/contract.rs
```diff
@@ -0,0 +1,299 @@
...
+                    .collect::<StdResult<Vec<_>>>()?;
+
+                Ok(Response::new()
+                    .add_attribute("method", "split")
+                    .add_messages(input_account_msgs))
+            }
+        }
+    }
+
+    // // Prepare transfer messages for each denom
```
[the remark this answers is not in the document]
stiiifff: fixed

contracts/services/reverse-splitter/README.md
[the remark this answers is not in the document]
stiiifff: done

contracts/services/reverse-splitter/src/contract.rs
```diff
@@ -0,0 +1,301 @@
...
+        params: &str,
+        denom: &CheckedDenom,
+    ) -> Result<Decimal, ServiceError> {
+        let denom_name = denom.to_string();
+        let res: DynamicRatioResponse = querier.query_wasm_smart(
+            contract_addr,
+            &DynamicRatioQueryMsg::DynamicRatio {
+                denoms: vec![denom_name.clone()],
+                params: params.to_string(),
+            },
```
Art3miX, line 250: Regarding querying a dynamic ration, we would probably need a wrapper of our own that will support this query because 99% other contracts wont.
Just something to note somewhere so we won't forgot.

contracts/services/reverse-splitter/src/msg.rs
```diff
@@ -0,0 +1,424 @@
...
+                params: params.to_string(),
+            },
+            None,
+        )
+    }
+
+    pub fn with_factor(mut self, factor: u64) -> Self {
+        self.factor = Some(factor);
+        self
+    }
```
Art3miX, line 176: The rest of the helper "with_" function are creating a new struct, I would suggest calling this as "set_factor" instead of with to make it clear.

contracts/services/reverse-splitter/src/msg.rs
```diff
@@ -0,0 +1,424 @@
...
+        }
+    }
+
+    // If there are ratios, we only allow an amount to be set for the base denom
+    if splits
+        .iter()
+        .any(|s| !matches!(s.amount, UncheckedSplitAmount::FixedAmount(..)))
+        && splits.iter().any(|s| {
+            matches!(s.amount, UncheckedSplitAmount::FixedAmount(..)) && s.denom != *base_denom
+        })
```
Art3miX, line 395: We already doing a loop above, we can have a flag set to true at the above loop if `any !UncheckedSplitAmount::FixedAmount(..)`, instead of doing another loop here.

contracts/services/splitter/Cargo.toml
```diff
@@ -29,11 +29,11 @@ valence-service-utils = { workspace = true }
...
-valence-account-utils      = { workspace = true }
-valence-service-utils      = { workspace = true, features = ["testing"] }
+cw-multi-test         = { workspace = true }
+cw-ownable            = { workspace = true }
+cw20                  = { workspace = true }
+cw20-base             = { workspace = true }
+sha2                  = { workspace = true }
+valence-account-utils = { workspace = true }
+valence-service-utils = { workspace = true, features = ["testing"] }
 valence-test-dynamic-ratio = { workspace = true }
```
Art3miX, line 39: Indentation

contracts/services/splitter/src/tests.rs
```diff
@@ -245,6 +246,46 @@ fn instantiate_fails_for_invalid_split_config() {
...
+    );
+    let cfg = suite.splitter_config(vec![split_cfg]);
+
+    // Instantiate Splitter contract
+    suite.splitter_init(&cfg);
+}
+
 #[test]
 #[should_panic(
     expected = "Configuration error: Duplicate split 'Native(\"untrn\")|AccountAddr(\"cosmwasm1ea6n0jqm0hj663khx7a5xklsmjgrazjp9vjeewejn84sanr0wgxq2p70xl\")' in split config."
```
Art3miX, line 291: I changed this line because I renamed an enum variant from `AccountAddr` to `Addr`, 
Would be great if we can avoid those `should_panic` when we need to include some parsing of enums or alike.

contracts/services/reverse-splitter/src/tests.rs
```diff
@@ -0,0 +1,907 @@
...
+const TEN_MILLION: u128 = 10_000_000_000_000_u128;
+const HUNDRED_MILLION: u128 = 100_000_000_000_000_u128;
+const HALF_MILLION: u128 = ONE_MILLION / 2;
+const FIVE_MILLION: u128 = TEN_MILLION / 2;
+
+#[derive(Getters, Setters)]
+struct ReverseSplitterTestSuite {
+    #[getset(get)]
+    inner: ServiceTestSuiteBase,
+    #[getset(get)]
```
Art3miX, line 31: What is the purpose of `getset(get)` again? you are building the get functions below anyways

contracts/services/reverse-splitter/src/tests.rs
```diff
@@ -0,0 +1,907 @@
...
+                    SplitAmount::FixedRatio(Decimal::percent(10u64)),
+                    None
+                )
+            ],
+            CheckedDenom::Native(NTRN.into())
+        )
+    );
+}
+
+// // Native & CW20 token amount splits
```
Art3miX, line 365: double //

Art3miX (approved):

contracts/services/splitter/Cargo.toml
stiiifff: fixed

contracts/services/reverse-splitter/src/msg.rs
stiiifff: ok

contracts/services/reverse-splitter/src/msg.rs
stiiifff: fixed

PR #53, opened 2024-10-04
Use ServiceAccountType for account addresses in ServiceConfig structs

stiiifff:


keyleu (approved):
Makes sense to me, will conflict with current token-swap e2e PR so let's merge this one first

PR #54, opened 2024-10-04
neutron e2e token swap test

keyleu:
Adds a test for a token swap on Neutron using our new system and local-ic.

For this, 2 base accounts are created and 2 splitters, one atomic action with 2 messages is created and sent through the authorization contract and queued on the processor. One tick on the processor executes the entire action atomically.

I found a small bug on the authorization contract regarding the json pointer (a "/" was missing at the start for them to work). Added also a unit-test to detect this in the future.
Re-organized local-ic folder a little to keep the utils there and refactored some code that is now being re-used by both our tests.
Also added the new test to CI.


.github/workflows/check-ci.yml
```diff
@@ -106,3 +106,47 @@ jobs:
...
+        run: |
+          cd local-interchaintest
+          local-ic start neutron_juno --api-port 42069 &
+          curl --head -X GET --retry 200 --retry-connrefused --retry-delay 5 http://localhost:42069
+
+      - name: Run polytone example
+        env:
+          RUST_LOG: debug
+        run: cargo run --package local-interchaintest --example token_swap
+
```
Art3miX, line 150: Rename

contracts/authorization/src/tests/processor.rs
```diff
@@ -1046,7 +1044,12 @@ fn queue_shifting_when_not_retriable() {
                                 message_type: MessageType::CosmwasmExecuteMsg,
                                 message: Message {
                                     name: "will_error".to_string(),
-                                    params_restrictions: None,
+                                    params_restrictions: Some(vec![
+                                        ParamRestriction::MustBeIncluded(vec![
+                                            "will_error".to_string(),
+                                            "error".to_string(),
+                                        ]),
+                                    ]),
```
Art3miX, line 1052: Can you explain? added restrictions but not added tests?

local-interchaintest/examples/token_swap.rs
```diff
@@ -0,0 +1,390 @@
...
+
+    info!("Swap authorization created!");
+    std::thread::sleep(std::time::Duration::from_secs(3));
+
+    info!("Send the messages to the authorization contract...");
+    let binary = Binary::from(
+        serde_json::to_vec(&valence_service_utils::msg::ExecuteMsg::<
+            ActionMsgs,
+            OptionalServiceConfig,
+        >::ProcessAction(ActionMsgs::Split {}))
```
Art3miX, line 331: Nit: you can use `ExecuteMsg::<_, Empty>` here or create a helper for simplicity, `ActionMsgs` is inferred here, and  `OptionalServiceConfig` is not used.

local-interchaintest/examples/token_swap.rs
```diff
@@ -0,0 +1,390 @@
...
+        .unwrap(),
+    );
+    let message = ProcessorMessage::CosmwasmExecuteMsg { msg: binary };
+
+    let send_msg = valence_authorization_utils::msg::ExecuteMsg::PermissionlessAction(
+        valence_authorization_utils::msg::PermissionlessMsg::SendMsgs {
+            label: "swap".to_string(),
+            messages: vec![message.clone(), message],
```
Art3miX, line 339: Why are 2 messages being sent?

Art3miX (approved):

contracts/authorization/src/tests/processor.rs
keyleu: the test is done right after this. I'm sending this message right after this so I confirm that the restriction is working (exactly this must be sent)

local-interchaintest/examples/token_swap.rs
keyleu: One for each splitter. The workflow is Account1->Splitter1->Account2 and Account2 ->Splitter2->Account1

local-interchaintest/examples/token_swap.rs
keyleu: Done!

.github/workflows/check-ci.yml
keyleu: done

PR #55, opened 2024-10-07
fix addresses in authorizations

Art3miX:
Changed to use `ServiceAccountType` in authorizations instead of string


keyleu (commented):
LGTM, there's a splitter test failing now, but all authorization/processor flow is fine

Art3miX:
Exactly why I hate the "expected error", I changed the naming of the enum variable from `AccountAddr` to `Addr`, so the error string was wrong.

keyleu (approved):

stiiifff (approved):

PR #61, opened 2024-10-10
Manager in e2e

Art3miX:
Big Pr includes:
1. implement the manager in our token swap e2e example
2. Fix constant 12 seconds sleep, we now query every second for the tx, for 15 seconds, erroring if the tx was not finalized after that time.
3. Manager config is global, which is nice because it allows us to freely use it, where we need it, and update it where we need it without the need to pass stuff from place to place, but it also adds a little complexity.

The biggest issue with the manager config is the fact we need to set the config differently based on where we are using it, which we would do anyways if the config was an argument the manager was taking, it just feels "hacky" with a global.
In production we would have files with all the data or take this data from a repo, in local-ic we need to get some data from the log file and some data after we upload contracts.

We also need to us tokio block_on, the reason is that the e2e tests are sync, and we cannot modify it to be async, so we have to use block_on in order to run our async functions in the manager.


contracts/authorization/src/tests/processor.rs
```diff
@@ -900,7 +901,8 @@ fn invalid_msg_rejected() {
             AtomicActionsConfigBuilder::new()
                 .with_action(
                     AtomicActionBuilder::new()
-                        .with_contract_address(&test_service_contract)
+                        .with_contract_address(ServiceAccountType::Addr(test_service_contract))
+                        // .with_contract_address(&test_service_contract)
```
keyleu, line 905: nit: leftover, can remove

local-interchaintest/examples/manager_test.rs
```diff
@@ -0,0 +1,115 @@
...
+    workflow_config.links.insert(
+        1,
+        Link {
+            input_accounts_id: vec![1],
+            output_accounts_id: vec![2],
+            service_id: 1,
+        },
+    );
+
+    // TODO: we need the id of the service here in contract_address
```
keyleu, line 89: what does this TODO mean here exactly?

local-interchaintest/src/utils/manager.rs
```diff
@@ -0,0 +1,210 @@
...
+        .unwrap();
+
+    // Update config with all the code ids
+    let mut code_ids_map = HashMap::new();
+    code_ids_map.insert(AUTHORIZATION_NAME.to_string(), authorization_code_id);
+    code_ids_map.insert(PROCESSOR_NAME.to_string(), processor_code_id);
+    code_ids_map.insert(BASE_ACCOUNT_NAME.to_string(), base_account_code_id);
+    code_ids_map.insert(SPLITTER_NAME.to_string(), splitter_code_id);
+    code_ids_map.insert(REVERSE_SPLITTER_NAME.to_string(), reverse_splitter_code_id);
+    code_ids_map.insert(FORWARDER_NAME.to_string(), forwarder_code_id);
```
keyleu, line 100: Can we pass this as an argument? Like have a struct something like:
```rust
pub struct ManagerSetupInfo {
       pub chain: String, // chain name
       pub contract_paths: Vec<String>, // path of all contracts we want to upload
}
```
and pass an array of this to this function? This way we can it up according to what we need for each test.
Right now it will only work for token_swap but would be nice that we can get it right from the start so that it can be reused for everything, including adding other domains.

We can still hard code the registry here to Neutron because we always need that.

This way each test can initialize the manager and tell it all the contracts that need to be uploaded and where

local-interchaintest/src/utils/manager.rs
```diff
@@ -0,0 +1,210 @@
...
+    // setup the chain info (and later bridge info) for the global config
+    // upload contracts to the chains and get their ids, and update the global config
+    // upload and instantiate a registry (for the manager) and update the global config
+
+    Ok(())
+}
+
+/// Get the chain infos and bridge info from the log file
+pub fn get_data_from_log() -> HashMap<String, ChainInfo> {
+    let log_file = fs::File::open("local-interchaintest/configs/logs.json")
```
keyleu, line 139: nit: add this path as a constant

local-interchaintest/src/utils/manager.rs
```diff
@@ -0,0 +1,210 @@
...
+                .get("chain_name")
+                .expect("Chain data must have chain_name")
+                .as_str()
+                .expect("chain_name must be a string")
+                .replace("local", "");
+
+            // TODO: We only want neutron for now, change later hardcoded values
+            if !chain_name.contains("neutron") {
+                return;
+            }
```
keyleu, line 164: can we already add all the chains? the CHAIN_NAMES we have are going to match with the values in logs so we could remove this

local-interchaintest/src/utils/manager.rs
```diff
@@ -0,0 +1,210 @@
...
+            let grpc = chain_data
+                .get("grpc_address")
+                .expect("Chain data must have grpc_address")
+                .as_str()
+                .expect("grpc_address must be a string");
+
+            chain_infos.insert(
+                NEUTRON_CHAIN_NAME.to_string(),
+                ChainInfo {
+                    name: NEUTRON_CHAIN_NAME.to_string(),
```
keyleu, line 180: we could use chain_name here then we can add all of them already

keyleu (commented):
Man this is dope!!! Love it, just a suggestion to not hard code the setup to Neutron for local-ic so that it can be reused by everyone without need to refactoring it

Art3miX:
Well, working with multiple chains will require the bridges to be set up as well, was specifically not going for multi-chain here because we need to add extra step, didn't want to optimize before we know how to.
Neutron is a special case, and in production we will us the actual chain names hardcoded into the conf files, local-ic gives us weird chain names like `localneutron-1` probably because you can spin multiple same chains, but this name will not work for us.

So in summary, this is just for a single chain set up, next we will do the multichain set up with setting up bridges (polytone) and the manager config for it.

local-interchaintest/examples/manager_test.rs
Art3miX: Also the manager_test is not an actual test we need to run, its just a helper for me to run some quick stuff avoiding any tests specific complexities.

local-interchaintest/src/utils/manager.rs
Art3miX: Will look into it in the multi-chain manager

local-interchaintest/src/utils/manager.rs
Art3miX: Neutron is a special case, and currently only wanted to set up everything for a single chain
Will look into it in the multi-chain manager

local-interchaintest/src/utils/manager.rs
Art3miX: Will look into it in the multi-chain manager

local-interchaintest/src/utils/manager.rs
keyleu: OK so for now (and my astroport e2e test I'll just hardcode the contracts here for now)

keyleu (approved):
LGTM, I'm planning to use this in my astroport e2e test as well, since I don't need multiple domains there. Planning on removing the leftovers now?

Art3miX:
Yup, forgot to push

PR #62, opened 2024-10-11
IBC transfer service - WIP

stiiifff:
Initial IBC transfer service implementation.

- ~~Only works from Neutron to other chain for now (uses Neutron SDK -specific types & messages to perform the IBC transfer)~~
- Fixed amount specified in config; effective amount transferred is configured amount minus IBC fees.
- Supports native tokens ~~& CW20 tokens (for chains supporting a custom port for CW20 transfers).~~
- ~~`RefundDust` operation allows to refund dust balance left on service back to the input account.~~
- E2E test with local-ic for Neutron->Juno and Juno->Neutron ✅

Todo:
- [x] Add unit tests to verify validation logic
- [x] Support non-Neutron CosmWasm chain
- [x] Add support for specifying a PFM map
- [x] Add README


local-interchaintest/examples/ibc_transfer_juno_ntrn.rs
```diff
@@ -0,0 +1,240 @@
...
+};
+use log::info;
+
+use valence_ibc_transfer_service::msg::{ActionsMsgs, ServiceConfig};
+use valence_service_utils::{denoms::UncheckedDenom, ServiceAccountType};
+
+fn main() -> Result<(), Box<dyn Error>> {
+    env_logger::init();
+
+    let ibc_denom = "ibc/4E41ED8F3DCAEA15F4D6ADC6EDD7C04A676160735C9710B904B7BF53525B56D6";
```
keyleu, line 28: this won't always work, channels are not consistent in local-ic so you might get different channels sometimes.
We have helpers to get the ibc denom after we have built the test_ctx. Examples:

```rust
let neutron_denom = test_ctx.get_native_denom().src(NEUTRON_CHAIN_NAME).get();
let neutron_on_juno_denom =  test_ctx
        .get_ibc_denom()
        .base_denom(neutron_denom)
        .src(NEUTRON_CHAIN_NAME)
        .dest(JUNO_CHAIN_NAME)
        .get();
```

local-interchaintest/examples/ibc_transfer_juno_ntrn.rs
```diff
@@ -0,0 +1,240 @@
...
+    let mut test_ctx = TestContextBuilder::default()
+        .with_unwrap_raw_logs(true)
+        .with_api_url(LOCAL_IC_API_URL)
+        .with_artifacts_dir(VALENCE_ARTIFACTS_PATH)
+        .with_chain(ConfigChainBuilder::default_neutron().build()?)
+        .with_chain(ConfigChainBuilder::default_gaia().build()?)
```
keyleu, line 34: nit, gaia is not being used, can be removed

local-interchaintest/examples/ibc_transfer_juno_ntrn.rs
```diff
@@ -0,0 +1,240 @@
...
+        .with_chain(ConfigChainBuilder::default_juno().build()?)
+        .with_log_file_path(LOGS_FILE_PATH)
+        .with_transfer_channels(NEUTRON_CHAIN_NAME, GAIA_CHAIN_NAME)
```
keyleu, line 37: nit: we don't need these channels, can remove this line

local-interchaintest/examples/ibc_transfer_juno_ntrn.rs
```diff
@@ -0,0 +1,240 @@
...
+            output_addr: output_account.clone(),
+            denom: UncheckedDenom::Native(ibc_denom.to_string()),
+            amount: transfer_amount.into(),
+            memo: "".to_owned(),
+            remote_chain_info: valence_ibc_transfer_service::msg::RemoteChainInfo {
+                channel_id: test_ctx
+                    .transfer_channel_ids
+                    .get(juno_ntrn_path)
+                    .unwrap()
+                    .clone(),
```
keyleu, line 154: nit: we have this syntax if you prefer to use this one:
```rust
test_ctx
                .get_transfer_channels()
                .src(JUNO_CHAIN_NAME)
                .dest(NEUTRON_CHAIN_NAME)
                .get()
```

which doesn't require unwrapping, cloning or defining the path

local-interchaintest/examples/ibc_transfer_ntrn_juno.rs
```diff
@@ -0,0 +1,238 @@
...
+};
+use log::info;
+
+use valence_ibc_transfer_service::msg::{ActionsMsgs, ServiceConfig};
+use valence_service_utils::{denoms::UncheckedDenom, ServiceAccountType};
+
+fn main() -> Result<(), Box<dyn Error>> {
+    env_logger::init();
+
+    let ibc_denom = "ibc/4E41ED8F3DCAEA15F4D6ADC6EDD7C04A676160735C9710B904B7BF53525B56D6";
```
keyleu, line 28: same as for the other test

local-interchaintest/examples/ibc_transfer_ntrn_juno.rs
```diff
@@ -0,0 +1,238 @@
...
+    let mut test_ctx = TestContextBuilder::default()
+        .with_unwrap_raw_logs(true)
+        .with_api_url(LOCAL_IC_API_URL)
+        .with_artifacts_dir(VALENCE_ARTIFACTS_PATH)
+        .with_chain(ConfigChainBuilder::default_neutron().build()?)
+        .with_chain(ConfigChainBuilder::default_gaia().build()?)
```
keyleu, line 34: same for other test, not needed

local-interchaintest/examples/ibc_transfer_ntrn_juno.rs
```diff
@@ -0,0 +1,238 @@
...
+            output_addr: output_account.clone(),
+            denom: UncheckedDenom::Native(NTRN_DENOM.to_string()),
+            amount: transfer_amount.into(),
+            memo: "".to_owned(),
+            remote_chain_info: valence_ibc_transfer_service::msg::RemoteChainInfo {
+                channel_id: test_ctx
+                    .transfer_channel_ids
+                    .get(ntrn_juno_path)
+                    .unwrap()
+                    .clone(),
```
keyleu, line 152: same like other test. Can build this directly with our localic-utils helpers

keyleu (changes requested):

contracts/services/ibc-transfer/src/contract.rs
```diff
@@ -0,0 +1,213 @@
...
+
+                let local_send_msg = SubMsg::new(
+                    execute_on_behalf_of(vec![transfer_msg], cfg.input_addr())?
+                        .change_custom()
+                        .ok_or_else(|| {
+                            ServiceError::ExecutionError(
+                                "Failed to change local send msg custom type.".to_owned(),
+                            )
+                        })?,
+                );
```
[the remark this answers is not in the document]
Art3miX: Each bridge might have its own rules, and encapsulating this in the service might be the best idea.
1 thing that comes with mind is how neutron ibc fee works, it is taken from the contract, or at least used to be taken from the sender contract.

On its own its not a big deal, but lets assume we have an amount sensitive workflow or service which requires for a specific amount to be handled.
If this service is working on `untrn` there will be an issue here, because the ibc fee will be taken from the account, which will result in `amount - 1`, in most cases this will be a big issue, but I think we should separate the "bridge payments" and the "working" funds of the workflow.

So for neutron IBC transfer, we should first check if the service have enough funds to pay for this fee before trying to send the funds.

contracts/services/generic-ibc-transfer/src/msg.rs
```diff
@@ -0,0 +1,100 @@
...
+    denoms::{CheckedDenom, UncheckedDenom},
+    error::ServiceError,
+    msg::ServiceConfigValidation,
+    ServiceAccountType, ServiceConfigInterface,
+};
+
+#[cw_serde]
+pub enum ActionsMsgs {
+    IbcTransfer {},
+}
```
keyleu, line 16: Can we rename to ActionMsgs? (Some services had ActionMsgs and others ActionMsgs so I renamed them all to ActionMsg), this way we can keep it consistent everywhere

contracts/services/neutron-ibc-transfer/src/msg.rs
```diff
@@ -0,0 +1,106 @@
...
+    denoms::{CheckedDenom, UncheckedDenom},
+    error::ServiceError,
+    msg::ServiceConfigValidation,
+    ServiceAccountType, ServiceConfigInterface,
+};
+
+#[cw_serde]
+pub enum ActionsMsgs {
+    IbcTransfer {},
+}
```
keyleu, line 17: Can we rename to ActionMsgs? (Some services had ActionMsgs and others ActionMsgs so I renamed them all to ActionMsg), this way we can keep it consistent everywhere

.github/workflows/check-ci.yml
```diff
@@ -179,3 +179,47 @@ jobs:
...
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Set up Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: ${{ env.GO_VERSION }}
+
+      - name: Install local-ic
+        run: git clone https://github.com/strangelove-ventures/interchaintest && cd interchaintest/local-interchain && make install
```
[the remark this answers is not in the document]
stiiifff: @keyleu actually I'm not sure what I have to change ..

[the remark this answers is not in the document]
keyleu: You can look at the one of the previous ones, remove the part where you install and add the part where you download the artifact and give execute permissions: 
e.g.:

- name: Download local-ic artifact
        uses: actions/download-artifact@v3
        with:
          name: local-ic
          path: /tmp
      - name: Make local-ic executable
        run: chmod +x /tmp/local-ic
      - name: Start local-ic and wait for it to be ready
        run: |
          cd local-interchaintest
          /tmp/local-ic start neutron_juno --api-port 42069 &
curl --head -X GET --retry 200 --retry-connrefused --retry-delay 5 http://localhost:42069

contracts/accounts/base_account/Cargo.toml
```diff
@@ -22,8 +22,7 @@ optimize = """docker run --rm -v "$(pwd)":/code \
 
 [dependencies]
 cosmwasm-schema       = { workspace = true }
-cosmwasm-std          = { workspace = true }
```
keyleu, line 25: Any reason we are adding cosmwasm1_4? If there isn't let's remove it or we won't be able to deploy on current mainnet Osmosis because they don't have this feature activated (due to devs messing up). But in general we shouldn't have features there we don't need.

contracts/services/generic-ibc-transfer/README.md
```diff
@@ -0,0 +1,78 @@
...
+  FullAmount,
+  // Transfer the specified amount of tokens.
+  FixedAmount(Uint128),
+}
+
+pub struct RemoteChainInfo {
+  // Channel of the IBC connection to be used.
+  channel_id: String,
+  // Port of  the IBC connection to be used.
+  port_id: Option<String>,
```
keyleu, line 64: nit: extra space in comment.

btw, do we need port_id here? Is there any scenario where it won't be "transfer"? For what I've seen in the contract it's not being used anywhere so anything you pass here will be ignored.

contracts/services/generic-ibc-transfer/src/msg.rs
```diff
@@ -0,0 +1,286 @@
...
+#[cw_serde]
+pub enum IbcTransferAmount {
+    FullAmount,
+    FixedAmount(Uint128),
+}
+
+#[cw_serde]
+pub struct RemoteChainInfo {
+    pub channel_id: String,
+    pub port_id: Option<String>,
```
keyleu, line 49: Extending on above, we shouldn't pass this parameter because it´s not being used.

contracts/services/neutron-ibc-transfer/README.md
```diff
@@ -0,0 +1,78 @@
...
+  FullAmount,
+  // Transfer the specified amount of tokens.
+  FixedAmount(Uint128),
+}
+
+pub struct RemoteChainInfo {
+  // Channel of the IBC connection to be used.
+  channel_id: String,
+  // Port of  the IBC connection to be used.
+  port_id: Option<String>,
```
keyleu, line 64: Similar to generic contract but one difference: here it's actually being used in the contract but are we even contemplating doing something different to ibc transfers? Looking at the contract we are using an MsgTransfer on both contracts so anything that is not "transfer" would not be valid.

I could see this making sense if we would have different kinds of IBC messaging supported here (e.g. IBC to another contract), but I think we would split that into a different service and using a base account with  ibc logic implemented (and therefore ibc ports, callbacks...)

contracts/services/generic-ibc-transfer/src/msg.rs
```diff
@@ -0,0 +1,286 @@
...
+                return Err(ServiceError::ConfigurationError(
+                    "Invalid IBC transfer config: remote_chain_info's ibc_transfer_timeout cannot be zero.".to_string(),
+                ));
+            }
+        }
+
+        if !self.denom_to_pfm_map.is_empty() {
+            return Err(ServiceError::ConfigurationError(
+                "Invalid IBC transfer config: specifying a denom_to_pfm_map on the Generic IBC transfer service is currently unsupported.".to_string(),
+            ));
```
keyleu, line 148: Why? I think this needs to be supported on both (since logic would be shared no need to test on both services though)

contracts/services/neutron-ibc-transfer/src/contract.rs
```diff
@@ -0,0 +1,141 @@
...
+                    }
+                };
+
+                // IBC Transfer funds from input account to output account on the remote chain
+                let block_time = env.block.time;
+                let ibc_send_msg = valence_ibc_utils::neutron::ibc_send_message(
+                    deps,
+                    env,
+                    cfg.remote_chain_info().channel_id.clone(),
+                    cfg.remote_chain_info().port_id.clone(),
```
keyleu, line 80: The way I see it, and extending on above, anything that is not "transfer" port here should not work as Neutron MsgTransfer is an IBC transfer would only work with port "transfer"

local-interchaintest/examples/ibc_transfer_juno_ntrn.rs
```diff
@@ -0,0 +1,305 @@
...
+    let start_input_balance = bank::get_balance(
+        test_ctx
+            .get_request_builder()
+            .get_request_builder(JUNO_CHAIN_NAME),
+        &input_account,
+    )
+    .iter()
+    .find(|bal| bal.denom == neutron_on_juno_denom)
+    .map_or(0, |bal| bal.amount.u128());
+    info!("Start input balance: {:?}", start_input_balance);
```
keyleu, line 99: nit: might be worth checking this right before the transfer execution as IBC transfers might take a bit longer (they depend on the relayer). This way you won't get 0 sometimes here, or just sleep for a bit longer (relayer is sometimes very slow on CI)

packages/ibc-utils/src/neutron.rs
```diff
@@ -0,0 +1,168 @@
...
+    to: String,
+    denom: String,
+    amount: u128,
+    memo: String,
+    timeout_height: Option<u64>,
+    timeout_timestamp: Option<u64>,
+    denom_to_pfm_map: BTreeMap<String, PacketForwardMiddlewareConfig>,
+) -> StdResult<CosmosMsg> {
+    // contract must pay for relaying of acknowledgements
+    // See more info here: https://docs.neutron.org/neutron/feerefunder/overview
```
keyleu, line 35: nit: contract is not paying anymore right, it's the base account now.

packages/ibc-utils/src/state.rs
keyleu, line 1: Is any of this still needed? My understanding is that we are not dealing with all these replies, sudo entry points so we can just remove this file?

packages/ibc-utils/src/state.rs
```diff
@@ -0,0 +1,68 @@
...
+pub const IBC_SUDO_ID_RANGE_SIZE: u64 = 1_000;
+pub const IBC_SUDO_ID_RANGE_END: u64 = IBC_SUDO_ID_RANGE_START + IBC_SUDO_ID_RANGE_SIZE;
+
+pub const REPLY_QUEUE_ID: Map<u64, Vec<u8>> = Map::new("reply_queue_id");
+
+const REPLY_ID: Item<u64> = Item::new("reply_id");
+
+/// SUDO_PAYLOAD - tmp storage for sudo handler payloads
+/// key (String, u64) - (channel_id, seq_id)
+/// every ibc chanel have its own sequence counter(autoincrement)
```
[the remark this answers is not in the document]
keyleu: typo still there, if we are keeping this file

contracts/accounts/base_account/Cargo.toml
```diff
@@ -22,8 +22,7 @@ optimize = """docker run --rm -v "$(pwd)":/code \
...
-
+cosmwasm-std          = { workspace = true, features = ["cosmwasm_1_4","stargate"] }
```
keyleu, line 25: Any reason we are adding cosmwasm1_4? If there isn't let's remove it or we won't be able to deploy on current mainnet Osmosis because they don't have this feature activated (due to devs messing up). But in general we shouldn't have features there we don't need.

.github/workflows/check-ci.yml
```diff
@@ -85,11 +85,6 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
-      - name: Set up Go
-        uses: actions/setup-go@v4
-        with:
-          go-version: ${{ env.GO_VERSION }}
-
```
keyleu, line 92: nice, ty for the cleanup

contracts/services/generic-ibc-transfer/README.md
stiiifff: It's used by the Neutron service (cfr. [here](https://github.com/timewave-computer/valence-protocol/blob/3be7979a51b5ba22553f287e85a43de21e48793d/packages/ibc-utils/src/neutron.rs#L52)).. and they share the same types. I wanted to leave some flexibility in the config.

contracts/services/generic-ibc-transfer/src/msg.rs
stiiifff: See my previous comment .. but I don't have a strong opinion about it. If you feel it's unnecessary, I can remove it.

contracts/services/neutron-ibc-transfer/README.md
stiiifff: I thought the we had a case for Covenant where we were IBC transferring CW20 tokens .. but the incoming port was `transfer` so it's still not needed in that case.

contracts/services/generic-ibc-transfer/README.md
keyleu: Yeah I know it's used but I think that any value we put there that is not "transfer" will not work. In fact not sure why Neutron is exposing it in their custom proto type, either they know something that I don't or this is redundant

local-interchaintest/examples/ibc_transfer_juno_ntrn.rs
stiiifff: So you mean the sleep just before that should be a bit longer, right ?

packages/ibc-utils/src/neutron.rs
stiiifff: Yeah .. but the account is a contract, right ? 😁

contracts/accounts/base_account/Cargo.toml
stiiifff: We are targetting the `cosmwasm1_4` feature of `cosmwasm_std` in the Workspace's cargo.toml.

.github/workflows/check-ci.yml
stiiifff: Yeah .. I figured it was not needed anymore.

contracts/accounts/base_account/Cargo.toml
keyleu: Yeah I think Ben removed it in his PR because it was giving trouble on Osmosis and we realized it wasn't being used anywhere

stiiifff: ah ok

packages/ibc-utils/src/state.rs
stiiifff: removed

keyleu (approved):

packages/ibc-utils/src/generic.rs
```diff
@@ -0,0 +1,43 @@
...
+            (Some(height), Some(timestamp)) => IbcTimeout::with_both(
+                IbcTimeoutBlock {
+                    revision: 2,
+                    height,
+                },
+                Timestamp::from_nanos(timestamp),
+            ),
+            _ => IbcTimeout::with_timestamp(env.block.time.plus_seconds(DEFAULT_TIMEOUT_TIMESTAMP)),
+        },
+        memo: Some(memo),
```
keyleu, line 41: quick question? cant we do the denom_to_pfm here?

stiiifff: The doc comment for that field says this:

> This field is only supported on chains with CosmWasm >= 2.0 and silently ignored on older chains.
> If you need support for both 1.x and 2.x chain with the same codebase, it is recommended to use `CosmosMsg::Stargate` with a custom MsgTransfer protobuf encoder instead.

keyleu: I see, makes sense

PR #63, opened 2024-10-14
Update logic into services

Art3miX:
Fairly straight forward change that abstract away the need for services to store raw config, this is done for them as long as they are using our methods, if not, they will need to manually calls the needed calls we are doing (which is not a lot):

The summary is:
1. Each service will have a "hidden" (abstracted) storage for the raw service config, and helpers for that can be found in: [packages/service-utils/src/raw_config.rs](https://github.com/timewave-computer/valence-protocol/compare/art3mix/service-update?expand=1#diff-59893be6405e3741306ee4923054fd6999c5cc031be1f73532cf62bf8d808f03)
2. The above is not magic, it is a regular Item storage for the raw service config, our helpers are using that storage to store and get the service config, on instantiation we call the `save_raw_service_config` which can be found here: [packages/service-base/src/lib.rs](https://github.com/timewave-computer/valence-protocol/compare/art3mix/service-update?expand=1#diff-22c2f6f9eac41e236e0d521610067bd8729b8f048fc102c1e5e7e86135c65463)
3. For update its a little bit more complex (not much), we created a trait so all optional service configs can implement, this is basically to expose a helper function: `OptionalServiceConfigTrait` can be found here:  [packages/service-utils/src/lib.rs](https://github.com/timewave-computer/valence-protocol/compare/art3mix/service-update?expand=1#diff-97df4c27d6c09bb58d2fca1d3897a19af778a0d9def52625607a372b95ca9e7f)
4. Our macro, implements this trait and the `update_raw` function, we need to remember that we are talking about `OptionalServiceConfig` here, so everything is an option, so what we do is, we get the saved `serviceConfig` that is in storage, we loop over all fields checking if it is Some, if it is, we are updating the `ServiceConfig` with the value in it, and save the it back to storage.
This allows us to update the raw service config regardless of what the checked config is.
5. All is left to do is to call this `update_raw` method to actually run it, we are doing it in the `updateConfig` execute message: [packages/service-base/src/lib.rs](https://github.com/timewave-computer/valence-protocol/compare/art3mix/service-update?expand=1#diff-22c2f6f9eac41e236e0d521610067bd8729b8f048fc102c1e5e7e86135c65463)
6. The last thing to do is to expose a query that we can query this raw `ServiceConfig`, for that I reused the extend macro we have in the covenant, so now a service can use a macro to get the full list of queries he must implement for a service to be a valence service, there is also a helper to get the raw `ServiceConfig` so no logic need to be done for the service programmer to implement them, can be as east as copy and paste. (probably can even be handled as a service utils helper)

One thing I didn't like is how each service must implement the same default for our queries, I hack I found that will allow us to set this default is by creating our own query function much like we have for execute, the `QueryMsg` we take from the service, we turn into Binary, and try to parse into our own `BaseQueryMsg` which will hold the queries we have, so if the service doesn't implement its own custom queries, and we can't parse the msg into our required queries, we fail, this is hackish but can provide a neat way for us to provide a default for service instead of them doing it by themselves.

Also in the macro I also added `ignore_optional`, so if a `ServiceConfig` has a field that we don't want to update, we can add this attribute on top of it, for example:
```rust
pub struct ServiceConfig {
  #[ignore_optional]
  pub ignore: String,
  pub include: String,
}
```
This will result in: 
```rust
pub struct OptionalServiceConfig {
  include: Option<String>
}
```

### DIFF
--- a/contracts/services/forwarder/src/contract.rs
+++ b/contracts/services/forwarder/src/contract.rs
@@ -6,7 +6,7 @@ use valence_service_utils::{
     msg::{ExecuteMsg, InstantiateMsg},
 };
 
-use crate::msg::{ActionsMsgs, Config, OptionalServiceConfig, QueryMsg, ServiceConfig};
+use crate::msg::{ActionsMsgs, Config, ServiceConfigUpdate, QueryMsg, ServiceConfig};
 
 // version info for migration info
 const CONTRACT_NAME: &str = env!("CARGO_PKG_NAME");
@@ -27,7 +27,7 @@ pub fn execute(
     deps: DepsMut,
     env: Env,
     info: MessageInfo,
-    msg: ExecuteMsg<ActionsMsgs, OptionalServiceConfig>,
+    msg: ExecuteMsg<ActionsMsgs, ServiceConfigUpdate>,
 ) -> Result<Response, ServiceError> {
     valence_service_base::execute(
         deps,
@@ -155,14 +155,14 @@ mod execute {
     use cosmwasm_std::{DepsMut, Env, MessageInfo};
     use valence_service_utils::error::ServiceError;
 
-    use crate::msg::{Config, OptionalServiceConfig};
+    use crate::msg::{Config, ServiceConfigUpdate};
 
     pub fn update_config(
         deps: &DepsMut,
         _env: Env,
         _info: MessageInfo,
         config: &mut Config,
-        new_config: OptionalServiceConfig,
+        new_config: ServiceConfigUpdate,
     ) -> Result<(), ServiceError> {
         new_config.update_config(deps, config)
     }
--- a/contracts/services/forwarder/src/contract.rs
+++ b/contracts/services/forwarder/src/contract.rs
@@ -183,9 +183,8 @@ pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> StdResult<Binary> {
             to_json_binary(&config)
         }
         QueryMsg::GetRawServiceConfig {} => {
-            let raw_config = valence_service_utils::raw_config::query_raw_service_config::<
-                ServiceConfig,
-            >(deps.storage)?;
+            let raw_config: ServiceConfig =
+                valence_service_utils::raw_config::query_raw_service_config(deps.storage)?;
             to_json_binary(&raw_config)
         }
     }
--- a/contracts/services/forwarder/src/contract.rs
+++ b/contracts/services/forwarder/src/contract.rs
@@ -19,7 +19,6 @@ pub fn instantiate(
     _info: MessageInfo,
     msg: InstantiateMsg<ServiceConfig>,
 ) -> Result<Response, ServiceError> {
-    // Custom
     valence_service_base::instantiate(deps, CONTRACT_NAME, CONTRACT_VERSION, msg)
 }
 
--- a/contracts/services/forwarder/src/contract.rs
+++ b/contracts/services/forwarder/src/contract.rs
@@ -155,16 +155,15 @@ mod execute {
     use cosmwasm_std::{DepsMut, Env, MessageInfo};
     use valence_service_utils::error::ServiceError;
 
-    use crate::msg::{Config, ServiceConfigUpdate};
+    use crate::msg::ServiceConfigUpdate;
 
     pub fn update_config(
         deps: DepsMut,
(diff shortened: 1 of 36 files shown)


contracts/services/forwarder/src/contract.rs
```diff
@@ -19,6 +19,7 @@ pub fn instantiate(
     _info: MessageInfo,
     msg: InstantiateMsg<ServiceConfig>,
 ) -> Result<Response, ServiceError> {
+    // Custom
```
keyleu, line 22: leftover?

contracts/services/template/src/msg.rs
```diff
@@ -53,4 +54,6 @@ impl OptionalServiceConfig {
 }
 
 #[cw_serde]
-pub struct Config {}
+pub struct Config {
+    pub admin: Addr,
+}
```
keyleu, line 59: why does Config need an admin now? I guess we can have an admin and an owner now and this admin is not updatable? I don't see this being used anywhere except the template now.

contracts/services/template/src/msg.rs
```diff
@@ -11,31 +11,32 @@ pub enum ActionsMsgs {
...
 
 #[cw_serde]
 #[derive(OptionalStruct)]
-pub struct ServiceConfig {}
+pub struct ServiceConfig {
+    /// We ignore this field when generating the OptionalServiceConfig
+    /// This means this field is not updatable
+    #[ignore_optional]
+    pub ignore_optional_admin: String,
+}
```
keyleu, line 28: Can you explain this a bit more?

contracts/services/forwarder/src/contract.rs
Art3miX: Yea, removed

keyleu (approved):

Art3miX:
Renamed some of the stuff to better names, also added `get_diff_update` in the macro, so you can query for the current `ServiceConfig` do changes to it, and get `ServiceConfigUpdate` with only the changes using this method.

stiiifff (approved):

stiiifff:
LGTM ! 🚀

contracts/services/template/src/msg.rs
```diff
@@ -1,56 +1,52 @@
...
     pub fn update_config(self, _deps: &DepsMut, _config: &mut Config) -> Result<(), ServiceError> {
         Ok(())
     }
 }
 
 #[cw_serde]
-pub struct Config {}
+pub struct Config {
+    pub admin: Addr,
+}
```
keyleu, line 52: should probably added the optional ones here too right? For it to make perfect sense to add other fields to update.

keyleu (approved):

stiiifff (approved):

keyleu (approved):

packages/valence-macros/src/helpers.rs
```diff
@@ -0,0 +1,61 @@
...
+pub(crate) fn has_skip_update_attr(attrs: &[Attribute]) -> bool {
+    attrs.iter().any(|attr| attr.path.is_ident("skip_update"))
+}
+
+pub(crate) fn get_option_inner_type(ty: &Type) -> Option<&Type> {
+    if let Type::Path(TypePath { path, .. }) = ty {
+        if let Some(segment) = path.segments.last() {
+            if segment.ident == "Option" {
+                if let PathArguments::AngleBracketed(args) = &segment.arguments {
+                    if let Some(GenericArgument::Type(inner_type)) = args.args.first() {
```
bekauz, line 18: (i'm not really experienced with `syn`) I wonder if something could go wrong on this?  if any the the `if let` checks above fail we return `None`, could it make sense to error instead?

bekauz (approved):
this looks great! just one question

packages/valence-macros/src/helpers.rs
bekauz: i highlighted this line because for `args.first()`, it's not really clear to me if we could be in a situation where `args` has more than one element somehow. I guess `Result<T, E>` would be two types? but not even sure how we can end up there..

Art3miX: Regarding the `args.first()`, we are looking into the args of `Option` which only has 1 arg possible, so i don't see a reason this will fail because before that we are making sure we are working with an `Option` type.

PR #71, opened 2024-10-15
remove NeutronMsg + wasm bindings and use protobuf directly

keyleu:
Move away from NeutronMsg and wasm bindings which ties the authorization contract to Neutron. With this change we can deploy the authorization contract to any tokenfactory chain and move away from what Neutron is deprecating at the moment.

Note: can't use neutron-std or latest commit of neutron-sdk because they inject cosmwasm_2_0 feature and we can't test with it atm. 
Once we can, we can replace neutron-sdk + cosmos-sdk-proto for neutron-std and Stargate for AnyMsg, added TODOs there. Code would be the same.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4737,6 +4737,7 @@ dependencies = [
 name = "valence-authorization"
 version = "0.1.0"
 dependencies = [
+ "cosmos-sdk-proto 0.20.0",
  "cosmwasm-schema 2.1.4",
  "cosmwasm-std 2.1.4",
  "cw-ownable",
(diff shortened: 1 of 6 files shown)


stiiifff (approved):

bekauz (approved):
makes sense to me. just one thing with regards to the future neutron crates version, would we be able to use it on osmosis?

keyleu:
> makes sense to me. just one thing with regards to the future neutron crates version, would we be able to use it on osmosis?

yes, they are using the same type_url

PR #72, opened 2024-10-15
osmo gamm lper

### LINKED ISSUE #66: osmo gamm liquidity provider

bekauz:
closes #66


contracts/services/osmosis-gamm-lper/src/balancer.rs
```diff
@@ -0,0 +1,219 @@
...
+        acceptable_spot_price_range.contains(pool_ratio)?;
+    }
+
+    // if the input balance is greater than the limit, we provision the limit amount.
+    // otherwise we provision the full input balance.
+    let provision_amount = if input_acc_asset_bal.amount > limit {
+        limit
+    } else {
+        input_acc_asset_bal.amount
+    };
```
keyleu, line 46: just realized I'm erroring when this happens on the astroport LPer so should probably change that to provide the limit instead of erroring.

Cargo.toml
```diff
@@ -83,3 +86,4 @@ hex                  = "0.4.3"
 margined-neutron-std = "4.2.0"
 neutron-test-tube    = "4.2.0"
 tokio                = "1.40.0"
+osmosis-test-tube    = { version = "25.0.0" }
```
keyleu, line 89: nit: keep same syntax

contracts/services/osmosis-gamm-lper/src/balancer.rs
```diff
@@ -0,0 +1,219 @@
...
+        let pool: Pool = pool_response
+            .pool
+            .ok_or_else(|| StdError::generic_err("failed to get pool"))?
+            .try_into()
+            .map_err(|_| StdError::generic_err("failed to decode proto"))?;
+
+        Ok(pool)
+    }
+}
+
```
keyleu, line 163: should prob be renamed right? You are querying the pool ratio and the pool

contracts/services/osmosis-gamm-lper/src/valence_service_integration.rs
keyleu, line 1: should we use the same syntax than rest of services? We're currently adding all this to msg.rs

packages/osmosis-utils/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+default = []
+testing = [
+    "dep:osmosis-test-tube",
+]
+
+[dependencies]
+cosmwasm-std            = { workspace = true }
+cosmwasm-schema         = { workspace = true }
+osmosis-test-tube       = { workspace = true, optional = true }
+cosmwasm-std-polytone   = { package = "cosmwasm-std", version = "1.5.7" }
```
keyleu, line 18: prob worth renaming this (I renamed it as well to cosmwasm-std-old) since it's not really related to polytone

contracts/services/osmosis-gamm-lper/src/balancer.rs
bekauz: renamed the `get_pool_response` to `query_pool`. as for the pool ratio, given that it's the balancer pool, the ratio is the spot price so felt like it describes the function a little better. do you think we should stick to pool ratio?

contracts/services/osmosis-gamm-lper/src/valence_service_integration.rs
bekauz: fair. it was a leftover from my previous branch where I had both gamm & cl logic.

keyleu (approved):

contracts/services/osmosis-gamm-lper/src/msg.rs
```diff
@@ -0,0 +1,177 @@
...
+use cw_ownable::cw_ownable_query;
+use osmosis_std::types::osmosis::{gamm::v1beta1::Pool, poolmanager::v1beta1::PoolmanagerQuerier};
+
+use valence_macros::OptionalStruct;
+use valence_osmosis_utils::utils::DecimalRange;
+use valence_service_utils::{
+    error::ServiceError, msg::ServiceConfigValidation, ServiceAccountType, ServiceConfigInterface,
+};
+#[cw_serde]
+pub enum ActionsMsgs {
```
keyleu, line 16: forgot this, In my other PR I renamed all to ActionMsgs (some had ActionsMsgs and others ActionMsgs). Can we rename it here too to have it consistent everywhere

PR #74, opened 2024-10-15
Fix astroport LPer limit handling

### LINKED ISSUE #73: fix single sided liquidity limit handling on astroport lper
Currently astroport lper errors when the account has more than the limit, but this is wrong. It should provide the limit instead.

keyleu:
Closes #73 

When looking at @bekauz Osmosis LPer PR I realized I was doing the handling of the limit wrong according to the spec. Modified it to LP the limit when the balance is higher instead of erroring.

### DIFF
--- a/contracts/services/astroport-lper/src/tests.rs
+++ b/contracts/services/astroport-lper/src/tests.rs
@@ -721,6 +721,7 @@ fn test_limit_single_sided_liquidity() {
 
     let input_acc_balance_after = query_balance(&setup.input_acc);
 
+    assert!(input_acc_balance_after > 0);
     assert_eq!(
         input_acc_balance_before - liquidity_provided,
         input_acc_balance_after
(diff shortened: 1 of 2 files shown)


stiiifff (approved):

contracts/services/astroport-lper/src/tests.rs
```diff
@@ -688,23 +688,42 @@ fn provide_single_sided_liquidity_cw20_lp_token() {
...
-            &ExecuteMsg::ProcessAction(ActionsMsgs::ProvideSingleSidedLiquidity {
-                asset: setup.inner.pool_asset1.clone(),
-                limit: Some(Uint128::new(1)),
-                expected_pool_ratio_range: None,
-            }),
-            &[],
-            setup.inner.processor_acc(),
-        )
-        .unwrap_err();
+    let query_balance = |address: &str| -> u128 {
```
Art3miX, line 693: Really should be a helper function somewhere

keyleu: changing it into the other PR since I have conflicts for this

PR #79, opened 2024-10-17
osmo CL lper + base account callbacks

bekauz:


keyleu:
Just looked at the callback handling for now, interesting, I wasn't thinking in using events but I guess it does the trick. The way I was thinking it was doing an ExecuteMsg to the original service in the reply of the base account and define a standard entry point for that in the ExecuteMsg of the service (we could add it with a macro or something).

Now I see the advantage of using events because it's quite straightforward and doesn't require an extra API entry point but I feel it can be dangerous because events might not be stable (?). Quoted from Simon a few days ago: "Events are never carefully designed. Tho whole stack from CosmWasm through Cosmos SDK to Comet BFT is an unstructured mess when it comes to events."

What if they decide to modify the events json structure in the future or they decide to modify the key like "wasm" into different kind of "wasm-". Or some chain out there running a fork and messing with these events, not sure just speculating. Then we wouldn't have it working everywhere.

Just speculating at this point not requiring any changes but prob worth a discussion

bekauz:
> Just looked at the callback handling for now, interesting, I wasn't thinking in using events but I guess it does the trick. The way I was thinking it was doing an ExecuteMsg to the original service in the reply of the base account and define a standard entry point for that in the ExecuteMsg of the service (we could add it with a macro or something).

Yeah, I just felt like that would complicate things on the account side a bit. 

> Now I see the advantage of using events because it's quite straightforward and doesn't require an extra API entry point but I feel it can be dangerous because events might not be stable (?). Quoted from Simon a few days ago: "Events are never carefully designed. Tho whole stack from CosmWasm through Cosmos SDK to Comet BFT is an unstructured mess when it comes to events."

100%, that's exactly the kind of worry I have and would love to hear more thoughts from others on this.

At least right now I feel like the way it's done is safe because I can't think of a situation where someone may "spoof" or mess with the expected event keys.
Account execution is gated by the approved services, so the account `reply` handler could be triggered iff the service initiates the submsg delegation.
On the service level, they await for that callback from the account. This feels like a closed loop where we control 2/3 (account + service, but not the target contract/module). I suppose the target address called by the underlying message we delegate to the account could do something malicious. Do you have any thoughts on this @keyleu ?

> What if they decide to modify the events json structure in the future or they decide to modify the key like "wasm" into different kind of "wasm-". Or some chain out there running a fork and messing with these events, not sure just speculating. Then we wouldn't have it working everywhere (would work for specific wasmd module versions)

Yeah, that's a good one. Ideally, imho, we would take advantage of the new `SubMsg` design (from cosmwasm 2.0) which has the [`payload: Binary`](https://github.com/CosmWasm/cosmwasm/blob/a3f0fa359fa68a786dec8f15554e86fca541d2e6/packages/std/src/results/submessages.rs#L36-L49) field. But given that it's not ready, I think transitioning to it will be easier from this attributes-based approach rather than from a dedicated execute msg?

bekauz:
One more thought on that semi-closed loop point i made where we do not control the integration, but control the rest.

I think it's a totally valid concern but I am not sure how far does it make sense to take it. We are already making a choice of trusting those integrations. In theory, there is nothing stopping some dApp from migrating their contracts into something malicious where the api would remain the same, but instead of performing the expected action, they would just take the input coins and send it to themselves?

keyleu:
> 100%, that's exactly the kind of worry I have and would love to hear more thoughts from others on this.
> 
> At least right now I feel like the way it's done is safe because I can't think of a situation where someone may "spoof" or mess with the expected event keys. Account execution is gated by the approved services, so the account `reply` handler could be triggered iff the service initiates the submsg delegation. On the service level, they await for that callback from the account. This feels like a closed loop where we control 2/3 (account + service, but not the target contract/module). I suppose the target address called by the underlying message we delegate to the account could do something malicious. Do you have any thoughts on this @keyleu ?

Yeah I'm not worried about chains messing with event keys but more of Cosmwasm transitioning to different even structures, but that's also very unlikely.

Yeah the account could get migrated and do malicious things but that can happen regardless with  anything, not particular to this, if the accounts are created with an admin and admin rights given to someone different than us.

This loop is closed so should be fine.

> Yeah, that's a good one. Ideally, imho, we would take advantage of the new `SubMsg` design (from cosmwasm 2.0) which has the [`payload: Binary`](https://github.com/CosmWasm/cosmwasm/blob/a3f0fa359fa68a786dec8f15554e86fca541d2e6/packages/std/src/results/submessages.rs#L36-L49) field. But given that it's not ready, I think transitioning to it will be easier from this attributes-based approach rather than from a dedicated execute msg?

Not sure how the payload would fix/change anything here tho or maybe I'm just not seeing it. The payload could be used if we added an entry point to the contract like I suggested then we could get the data from the payload instead of the API msg.

E.g. We could have in the service:
ValenceCallback {} and take it from the payload instead of ValenceCallback{ callback_msg: Binary }.

Or maybe I'm just not seeing it.

bekauz:
> > Yeah, that's a good one. Ideally, imho, we would take advantage of the new `SubMsg` design (from cosmwasm 2.0) which has the [`payload: Binary`](https://github.com/CosmWasm/cosmwasm/blob/a3f0fa359fa68a786dec8f15554e86fca541d2e6/packages/std/src/results/submessages.rs#L36-L49) field. But given that it's not ready, I think transitioning to it will be easier from this attributes-based approach rather than from a dedicated execute msg?
> 
> Not sure how the payload would fix/change anything here tho or maybe I'm just not seeing it. The payload could be used if we added an entry point to the contract like I suggested then we could get the data from the payload instead of the API msg.

I was thinking something along the lines of receiving the callback in the account, taking the reply, wrapping it in this payload, and firing it back to the calling service. Then service gets the reply and ignores everything except the payload, because that payload should be indistinguishable from as if the service fired that message directly. Msg and the reply id fired by the account are configured by the service, so the service can just treat it as if the account didn't exist.

keyleu:
> firing it back to the calling service.

Fire it back how?, only see the option of an ExecuteMsg so we're back to needing an entry point. Because the Reply we're getting on the service will only have the events of the wasm execution (what you have now)

bekauz:
yeah kind of dropped the idea of complicating this further for now @keyleu

stiiifff (commented):
Looks super neat, and the addition of callbacks on Valence accounts opens up more use cases.
Maybe one thing: it would be nice to add one or more basic tests for the callbacks on the account (in `contracts\accounts\base_account\src\tests.rs`).

bekauz:
@stiiifff done!

contracts/services/osmosis-cl-lper/src/contract.rs
```diff
@@ -0,0 +1,232 @@
...
+
+    // we delegate the position creation message to the input account
+    let delegated_input_acc_msgs = execute_submsgs_on_behalf_of(
+        // we expect a reply from this submsg so we pass it as a submessage
+        vec![SubMsg::reply_always(create_cl_position_msg, REPLY_ID)],
+        // associate this msg with a cfg payload which will be used in the reply
+        // to restore the state used during this function
+        Some(to_json_string(&cfg)?),
+        &cfg.input_addr.clone(),
+    )?;
```
keyleu, line 182: Why reply_always here?. Unless in the base account you want to save something when it errors (don't think it's the case as they don't store state afaik) this should be a reply_on_success. This way if the CL position fails we just stop

keyleu (commented):
Looks great to me, just 1 comment regarding the handling of the reply which shouldn't affect much but we could stop execution earlier.
Also might be worth to use newest test-tube (released today) as I guess you'll be able to get rid of using old cosmwasm_std for example

contracts/services/osmosis-cl-lper/src/contract.rs
bekauz: good one; I had it there from before accounts had callbacks and I was clearing the temp state on both success & failure but this is indeed cleaner now

keyleu (approved):

packages/osmosis-utils/src/utils.rs
```diff
@@ -101,16 +101,31 @@ pub mod cl_utils {
...
+            // calculating the lower bound of the current tick range works as follows:
+            // 1. Divide the current tick by the tick spacing (using euclidean division).
+            //  Euclidian division is used here to cover cases where the current tick is
+            //  negative. Regular integer division rounds towards 0, which would return an
+            //  off-by-one bucket (towards positive values).
+            // 2. Multiply the result by the tick spacing
+            let lower_bound = current_tick
+                .checked_div_euclid(tick_spacing)
+                .map_err(|e| StdError::generic_err(e.to_string()))?
+                .checked_mul(tick_spacing)?;
```
stiiifff, line 123: Yeah, we should absolutely always follow the rule _no unchecked math_

stiiifff:
LGTM

stiiifff (approved):

PR #82, opened 2024-10-23
Review unchecked maths in codebase

stiiifff:


contracts/services/astroport-lper/src/astroport_cw20.rs
```diff
@@ -85,7 +85,8 @@ fn create_xyk_liquidity_msg(
     let halved_coin = cosmwasm_std::Coin {
         denom: asset_balance.denom.clone(),
         amount: cosmwasm_std::Uint128::from(asset_balance.amount.u128())
-            / cosmwasm_std::Uint128::from(2u128),
+            .checked_div(cosmwasm_std::Uint128::from(2u128))
+            .expect("denominator is not zero; qed"),
```
bekauz, line 89: does this make sense when the denominator is hardcoded to 2? also can't we use `?` to unpack the result of division?

contracts/services/astroport-lper/src/astroport_native.rs
```diff
@@ -86,7 +86,8 @@ fn create_xyk_liquidity_msg(
     let halved_coin = cosmwasm_std::Coin {
         denom: asset_balance.denom.clone(),
         amount: cosmwasm_std::Uint128::from(asset_balance.amount.u128())
-            / cosmwasm_std::Uint128::from(2u128),
+            .checked_div(cosmwasm_std::Uint128::from(2u128))
+            .expect("denominator is not zero; qed"),
```
bekauz, line 90: same as above

bekauz (approved):
makes sense, just curious why `.expect()` over `?` to unpack? cw math libraries return `DivideByZero` error in these cases so I think it would get propagated fine with `?`

contracts/services/astroport-lper/src/astroport_cw20.rs
stiiifff: It's a common way of making more explicit invariants that you know to be true / infaillible (I've seen this used in many codebases) i.e. it's clearer to use an `.expect` with a message than just an `.unwrap`.

bekauz: Only thing a `checked_div` checks here is for division by zero, and we are dividing by a hardcoded value (2) so I'm not sure if it's any cleaner. Imho it's just extra operations.

Regarding `.expect()`, totally agree that we shouldn't simply `.unwrap()`. Personally I see `.expect()` as very similar to unwrap in that it still panics, just with a custom message. In contrast with `?`, which propagates the underlying error message. I think that propagating underlying errors, especially for stuff like cosmwasm math, is the correct approach because we would always receive the same uniform error messages across our codebase.
Another point is regarding consistency, - use `?` much more than `.expect()`.

Regardless, my approval is already here as the way it's currently done would (ofc) work. Just nitpicking about uniformity.

PR #83, opened 2024-10-23
PFM support in Generic IBC transfer service

stiiifff:


local-interchaintest/examples/ibc_transfer_osmo_juno_gaia_pfm.rs
```diff
@@ -0,0 +1,296 @@
...
+                        .get_transfer_channels()
+                        .src(OSMOSIS_CHAIN_NAME)
+                        .dest(JUNO_CHAIN_NAME)
+                        .get(),
+                    hop_to_destination_chain_channel_id: test_ctx
+                        .get_transfer_channels()
+                        .src(JUNO_CHAIN_NAME)
+                        .dest(GAIA_CHAIN_NAME)
+                        .get(),
+                    hop_chain_receiver_address: JUNO_CHAIN_ADMIN_ADDR.to_string(),
```
bekauz, line 212: for debugging I'd start with setting this to an invalid bech32 address so that the funds don't settle in this addr in case of a failed pfm

stiiifff: Test is now passing .. it was due to the (incorrect) use of an old version (non-cosmwasm 2.0) Osmosis chain.

PR #86, opened 2024-11-05
Fix neutron-ibc-transfer-service fee handling

stiiifff:


keyleu (dismissed):

keyleu:
[send_request payload]: {"action":"query","chain_id":"localneutron-1","cmd":"bank balances neutron14hj2tavq8fpesdwxxcu44rty3hh90vhujrvcmstl4zr3txmfvw9s5c2epq --node=%RPC% --chain-id=%CHAIN_ID% --output=json"}
thread 'main' panicked at local-interchaintest/examples/ibc_transfer_ntrn_juno_gaia_pfm.rs:283:5:
assertion `left == right` failed
  left: 0
 right: 2000
 
 PFM test is failing, is this related to the other PR?

stiiifff:
@keyleu Fixed and tweaked the behavior as follows:
- If the denom being transferred is NTRN:
  - If the full sender's balance is being transferred, the IBC fee is deducted from the account's balance.
  - Otherwise, verify that the account's balance is sufficient to transfer the requested amount + the fee amount (fail otherwise).
- If the denom being transferred is not NTRN, just ensure that the sender account's balance has enough NTRN to cover for the IBC fees.
/cc @bekauz

bekauz:
I think in case denom being transferred is ntrn, all we need to assert is that the amount is gte than the minimum fee. This is slightly annoying because if this is being done on neutron, we query that fee on demand. The non-ntrn transfer case makes sense to me! will give this a look shortly

packages/ibc-utils/src/neutron.rs
```diff
@@ -38,19 +41,45 @@ pub fn ibc_send_message(
...
-        .checked_sub(total_fee.u128())
-        .ok_or_else(|| StdError::generic_err("Amount too low to pay for IBC transfer fees."))?;
+
+    // Sender's account balance for fee denom (NTRN)
+    let sender_ntrn_balance = deps.querier.query_balance(sender, NTRN_DENOM)?.amount;
+
+    let transfer_amount = if denom.to_string() == NTRN_DENOM {
+        if Uint128::from(amount) == sender_ntrn_balance {
+            // The full balance is being transferred .. deduct the fee from the transfer amount
+            Uint128::from(amount).checked_sub(total_fee)?
```
bekauz, line 51: I don't think this deduction is necessary, it gets "taken" by the ibc module automatically. Or is this meant to reserve the fees for an additional transfer in the future?

packages/ibc-utils/src/neutron.rs
```diff
@@ -38,19 +41,45 @@ pub fn ibc_send_message(
...
-    let msg = match denom_to_pfm_map.get(&denom) {
+    let msg = match denom_to_pfm_map.get(&denom.to_string()) {
         None => neutron_sdk::proto_types::neutron::transfer::MsgTransfer {
             source_port: "transfer".to_string(),
             source_channel: channel.clone(),
-            sender,
+            sender: sender.to_string(),
             receiver: to.clone(),
             token: Some(coin),
             timeout_height: Some(cosmos_sdk_proto::ibc::core::client::v1::Height {
```
bekauz, line 85: I know this is not the part of this PR but are we sure we want to use height-based timeouts? I don't think hardcoding revision number to 2 is right

packages/ibc-utils/src/neutron.rs
Art3miX: AFAIK it needs to know the amount to send, but it can't send the full amount if the denom is NTRN, so we need to reduce the fee to get the amount we can send.

Art3miX (approved):

contracts/services/neutron-ibc-transfer/src/tests.rs
```diff
@@ -150,6 +161,86 @@ impl ServiceTestSuite<Empty, NeutronQuery> for IbcTransferTestSuite {
...
+    }
+}
+
+impl Module for NeutronKeeper {
+    type ExecT = Empty;
+    type QueryT = NeutronQuery;
+    type SudoT = SudoMsg;
+
+    /// Currently we only implement register ICA and ibcTransfer and SubmitTx,
+    /// maybe we should implement other stuff as well?
```
keyleu, line 183: left over comment from somewhere else?

contracts/services/neutron-ibc-transfer/src/tests.rs
```diff
@@ -150,6 +161,86 @@ impl ServiceTestSuite<Empty, NeutronQuery> for IbcTransferTestSuite {
...
+impl NeutronKeeper {
+    pub fn new() -> Self {
+        Self {}
+    }
+}
+
+impl Module for NeutronKeeper {
+    type ExecT = Empty;
+    type QueryT = NeutronQuery;
+    type SudoT = SudoMsg;
```
keyleu, line 180: Maybe it's worth moving this to somewhere else so that it can be used for any Neutron specific contracts instead of having to copy it

keyleu (approved):
LGTM, lmk if you want to merge this first before me merging renames

stiiifff:
> I think in case denom being transferred is ntrn, and we are transferring the full available balance then all we need to assert is that the amount is gte than the minimum fee. It will just get subtracted from the total transfer amount and yield the same result. Not 100% sure about what is the case for non-full balance transfers but I suspect that it would get taken from the account balance that is not being transferred. Would be nice to double check this. The non-ntrn transfer case makes sense to me! will give this a look shortly

I have re-checked and .. when transferring the full amount, if we don't deduct the fees, the transfer fails .. so it looks like the fees are not automatically deducted. Re. the difference between a fixed amount and the full balance:

- Fixed amount: transfer that amount precisely, and ensure that the balance is sufficient to also cover for IBC fees, fail otherwise.
- Full balance: transfer whatever is remaining on the account. Only fail if the balance is below the fee amount.

packages/ibc-utils/src/neutron.rs
stiiifff: @bekauz hardcoded value was wrong, my bad. And I agree with your remark .. height-based timeouts are a bad idea to deal with, as they require some understanding about the underlying infrastructure (e.g. client revision number). Time-based timeouts provide a simpler (and more stable) abstraction. I have removed the ability to specify height-based timeouts, let me know what you think.

packages/ibc-utils/src/generic.rs
```diff
@@ -15,28 +15,18 @@ pub fn ibc_send_message(
...
-            (Some(height), Some(timestamp)) => IbcTimeout::with_both(
-                IbcTimeoutBlock {
-                    revision: 2,
-                    height,
-                },
-                Timestamp::from_nanos(timestamp),
-            ),
-            _ => IbcTimeout::with_timestamp(env.block.time.plus_seconds(DEFAULT_TIMEOUT_TIMESTAMP)),
+        timeout: match timeout_timestamp {
+            Some(timestamp) => IbcTimeout::with_timestamp(Timestamp::from_nanos(timestamp)),
```
bekauz, line 26: does this need to be expressed in nanos? the default timestamp const is expressed in seconds (600). also feels like passing the values for this arg (`Option<u64>`) would mean passing stuff like `60000000000` to express a minute long timeout. feels like seconds are expressive enough, no?

bekauz (approved):
just one comment regarding the timeout units

packages/ibc-utils/src/generic.rs
keyleu: missed this one, yeah I think it should def be seconds here

contracts/services/neutron-ibc-transfer/src/tests.rs
stiiifff: I suggest to do this later, when the need arises .. I don't want to delay the merge of this PR.

packages/ibc-utils/src/neutron.rs
```diff
@@ -80,12 +80,11 @@ pub fn ibc_send_message(
...
-                env.block
-                    .time
-                    .plus_seconds(DEFAULT_TIMEOUT_TIMESTAMP)
-                    .nanos(),
-            ),
+            timeout_timestamp: env
+                .block
+                .time
+                .plus_seconds(timeout_seconds.unwrap_or(DEFAULT_TIMEOUT_TIMESTAMP))
+                .nanos(),
```
keyleu, line 87: I think this is incorrect and why an overflow is happening right? Should this be nanos? In the generic transfer you are not converting to nanos

packages/ibc-utils/src/neutron.rs
```diff
@@ -97,12 +96,11 @@ pub fn ibc_send_message(
...
-                    env.block
-                        .time
-                        .plus_seconds(DEFAULT_TIMEOUT_TIMESTAMP)
-                        .nanos(),
-                ),
+                timeout_timestamp: env
+                    .block
+                    .time
+                    .plus_seconds(timeout_seconds.unwrap_or(DEFAULT_TIMEOUT_TIMESTAMP))
+                    .nanos(),
```
keyleu, line 103: same as above

packages/ibc-utils/src/neutron.rs
keyleu: nevermind checked the spec and it seems it's correctly in nanos

keyleu:
looking at the previous contract code and this one.

You're doing in contract.rs 
`rust
let ibc_send_msg = valence_ibc_utils::generic::ibc_send_message(
                    env,
                    cfg.remote_chain_info().channel_id.clone(),
                    cfg.output_addr().to_string(),
                    cfg.denom().to_string(),
                    amount.u128(),
                    cfg.memo().clone(),
                    cfg.remote_chain_info()
                        .ibc_transfer_timeout
                        .map(|timeout| block_time.plus_seconds(timeout.u64()).nanos()),
                    cfg.denom_to_pfm_map().clone(),
                )
.map_err(|err| ServiceError::ExecutionError(err.to_string()))?;`

but then again in the ibc_send_message function you are adding the block time plus the timestamp again.

So you are adding it twice and when you are passing it  to the function since you are passing nanos you are overflowing u64

keyleu (changes requested):
needs to fix the  double timestamp addition and use nanos on both generic/neutron version

keyleu (approved):

PR #90, opened 2024-11-06
Manager partial update

Art3miX:
Allow updating service configs and authorizations for a workflow.

It accepts what updates the user wants to do, and output a list of `WasmMsg::ExecuteMsg` that needs to be executed by the owner in order to fulfil those changes.

The full process can be viewed in the `manager_test`.


keyleu (approved):
LGTM.

One question, we're not allowing to add new accounts with the update I guess right? In that case, if someone updates the workflow with new input/output accounts they will not be part of the registry and he needs to maintain that?

Also, want me to wait for this to be merged before I merge rename PR?

Art3miX:
Yea, this partial update doesn't allow of modifying the program diagram (adding accounts or services), it is there in case you need some authorizations added/modified/disabled/enabled, and fix library configs that might be outdated or were set wrong. 
If you need more then that, you need to create a new program.

Regarding merging, whatever conflict will be easier to handle, I don't think it matters much, so whatever is easier to fix.

workflow-manager/src/helpers.rs
```diff
@@ -32,3 +32,8 @@ pub fn addr_humanize(prefix: &str, canonical: &CanonicalAddr) -> Result<String,
     let prefix = Hrp::parse(prefix).map_err(|_| anyhow!("Invalid bech32 prefix"))?;
     encode::<Bech32>(prefix, canonical.as_slice()).map_err(|_| anyhow!("Bech32 encoding error"))
 }
+
+// /// Helper for serde default for bool
+// pub(crate) fn bool_true_default() -> bool {
+//     true
+// }
```
stiiifff, line 39: cleanup ?

workflow-manager/src/workflow_update.rs
```diff
@@ -0,0 +1,284 @@
...
+impl WorkflowConfigUpdate {
+    /// Modify an existing config with a new config
+    pub async fn update(&mut self, connectors: &Connectors) -> ManagerResult<UpdateResponse> {
+        let neutron_domain = Domain::CosmosCosmwasm(NEUTRON_CHAIN.to_string());
+
+        // get the old workflow config from registry
+        let mut neutron_connector = connectors.get_or_create_connector(&neutron_domain).await?;
+
+        if self.id == 0 {
+            return Err(ManagerError::IdIsZero);
```
stiiifff, line 69: A bit of a weird error .. Maybe you could create a constructor where you enforce that it's not possible to create WorkflowConfigUpdate with an `id` equal to zero (and any other basic validation).

workflow-manager/src/workflow_update.rs
```diff
@@ -0,0 +1,284 @@
...
+        }
+
+        let mut config = neutron_connector.get_workflow_config(self.id).await?;
+
+        let mut instructions: VecDeque<CosmosMsg> = VecDeque::new();
+        let mut new_authorizations: Vec<AuthorizationInfo> = vec![];
+
+        if let Some(new_owner) = self.owner.clone() {
```
stiiifff, line 77: This could be modelled as a `ChangeOwner` command / instruction.

workflow-manager/src/workflow_update.rs
```diff
@@ -0,0 +1,284 @@
...
+                .build();
+            let authorization_builder = AuthorizationBuilder::new()
+                .with_label(&label)
+                .with_mode(AuthorizationModeInfo::Permissioned(
+                    valence_authorization_utils::authorization::PermissionTypeInfo::WithoutCallLimit(vec![config.owner.clone()]),
+                ))
+                .with_priority(Priority::High)
+                .with_actions_config(actions_config);
+
+            new_authorizations.push(authorization_builder.build());
```
stiiifff, line 134: Should you check if it doesn't exist yet before adding it ?

workflow-manager/src/workflow_update.rs
```diff
@@ -0,0 +1,284 @@
...
+                )
+                .unwrap(),
+                funds: vec![],
+            }
+            .into(),
+        );
+
+        // Save the new config to the registry
+        neutron_connector.update_workflow_config(config).await?;
+
```
stiiifff, line 257: General comment: there's a lot happening in this function, and those are potentially sensitive operations (as in, breaking) on all aspects of a workflow/program config. I think it would be good (for future maintenance's sake) to organize the logic in a clear, staged structure .. as the current logic is mixing updating the authorizations, creating instructions, .. it's easy to get lost in what's happening.

workflow-manager/src/workflow_update.rs
Art3miX: Not sure what you mean.

workflow-manager/src/workflow_update.rs
Art3miX: Good point

workflow-manager/src/workflow_update.rs
Art3miX: It might not be clear what update does.

Update should not break the program, what it does is it allows updating library configs and authorizations, thats it, the manager cannot execute anything, and this is what instructions are, the update function output the messages that needs to be executed.

so there are 3 things happening here:
1. We build the messages to add/modify/disable/enable **authorizations**
2. We create the message to update the **library config**
3. Only the processor can update library configs because its the owner of the library contract, so we need to build the authorization that allows this operation.

After the update method ran, it output what messages needs to be executed by the owner in order to perform all those changes.

workflow-manager/src/workflow_update.rs
stiiifff: I mean that you are deriving whether to create a Change owner instruction from an optional field. The intent of modifying the owner could be modelled directly as a `ChangeOwner` instruction / command.

workflow-manager/src/workflow_update.rs
stiiifff: Maybe something more explicit like `InvalidWorkflowId` ?

workflow-manager/src/workflow_update.rs
stiiifff: I understand .. but as you mentioned, there are 3 things happening .. and the logic for those 3 things is interweaved / inter-related. I just suggest that making that code a bit clearer would help with future maintenance.

Art3miX: Added comments to explain a little more whats happeneing

program-manager/src/program_update.rs
```diff
@@ -0,0 +1,303 @@
...
+        // Save the updated config to the registry
+        neutron_connector.update_program_config(config).await?;
+
+        Ok(UpdateResponse {
+            instructions: instructions.into(),
+        })
+    }
+}
+
+fn verify_authorization_not_exists(
```
stiiifff, line 283: This feels redundant ..

stiiifff (approved):

program-manager/src/program_update.rs
Art3miX: Yea, but explicit in what we verify, and we want to error here, so I felt like it will be better encapsulated in a helper function instead of bool and errors in code it self.
Should I change it?

PR #92, opened 2024-11-07
Docs update

keyleu:
Extended docs with new terminology and focusing on Authorization/Processor. They will be further extended with Program manager and services.


docs/src/components/accounts_and_libraries.md
```diff
@@ -0,0 +1,23 @@
+# Accounts and Libraries
+
+- **Valence Account:** An escrow contract that can hold balances for various supported token-types (e.g., in Cosmos `ics-20` or `cw-20`) . Accounts register valence libraries that can perform token operations on the account.
+
+- **Valence Libraries:** Valence libraries are registered with Valence accounts. Accounts are either registered as input accounts or output accounts. Libraries perform token operations on the input accounts, such as splitting, routing, or providing liquidity on a DEX, and depending on the business logic in the Library, they route the resulting tokens to the output accounts. Libraries expose _Functions_ that can be performed on them.
```
Art3miX, line 5: I think we need to refrain from using input and output accounts, especially in defining libraries.

"Libraries executes a function on accounts"

The fact this function usually or might take funds from account A and move them to account B, doesn't mean its what define libraries.

Good example is the rebalancer library, it has no "output account", it executes a "register" function on an account and that's it.  

Just like programming libraries, you approve them by importing them, which grants them the ability to execute functions where you imported them.

docs/src/components/programs.md
```diff
@@ -0,0 +1,11 @@
+# Programs
+
+- **Valence Program:** A program is an instance of the Valence Protocol. It is a particular arrangement and configuration of _ValenceAccounts_ and _Function_ across multiple _domains_. Programs are also associated with a set of _Subroutines_ that can be executed on the _Program_. For example, a POL lending relationship between two parties may be set up as a program.
```
Art3miX, line 3: Shouldn't it be: `It is a particular arrangement and configuration of _ValenceAccounts_ and _Libraries_ across multiple _domains_`

docs/src/components/programs.md
```diff
@@ -0,0 +1,11 @@
...
+
+- **Domain:** Environments where _ValenceAccounts_ or _Libraries_ can be instantiated. Domains are defined by three properties:
+  1. Chain _(e.g., Neutron, Osmosis, Ethereum mainnet)_
+  2. Execution environment _(e.g., CosmWasm, EVM, SVM)_
+  3. Bridge from main domain _(e.g., Polytone over IBC, Hyperlane)_
```
Art3miX, line 8: I think bridge should not be part of the domain definition.

Main reason is that bridge is a list of open channels of communications that a domain has, a domain can have 2 channel of communication (polytone and hyperlane), so other domains can talk with this domain over those 2 channels.
It doesn't mean we have 2 different domains, rather 1 domain with multiple channels.

Art3miX (commented):
What is the defined goal of this doc? 
Is it for developers? just for us?

keyleu:
> What is the defined goal of this doc? Is it for developers? just for us?

For both

docs/src/components/programs.md
keyleu: Correct, fixed it.

docs/src/components/accounts_and_libraries.md
keyleu: Agree that once we move to singleton libraries we probably need to rephrase this to better represent how services / accounts are linked.

docs/src/components/programs.md
keyleu: I'm still confused by this, on some conversations we had we included the Bridge in the domain in others we didn't. The truth is right now we are including the bridging configuration when we add an external domain to the authorization contract. So if for example we add Osmosis with Hyperlane and Osmosis with IBC they would be represented as different external domains. Prob need more clarification on this @uditvira

stiiifff (approved):

PR #100, opened 2024-11-15
CI improvements: rust-cache, free-disk-space

stiiifff:


.github/workflows/check-ci.yml
```diff
@@ -21,6 +21,12 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      - name: Cache Build artefacts
+        uses: Swatinem/rust-cache@v2.7.5
+        with:
+          cache-on-failure: true
+          shared-key: ${{ github.event.pull_request.number || github.ref }}
+
```
keyleu, line 29: nice, this caches the ./target folder so when rerunning failed tests it's much faster right?

keyleu (approved):

.github/workflows/check-ci.yml
stiiifff: yep !

PR #101, opened 2024-11-16
json schema ci checks

### LINKED ISSUE #95: automate schema validation checks in CI

bekauz:
closes #95 

works by generating the schemas in the CI, and asserting that it results in no changes to the file tree.
this way, if any of our contracts api in a given pr had changed, but schemas were not regenerated, the assertion will fail because the up to date schema generated in the job will result in a diff.

### DIFF
--- a/.github/workflows/check-ci.yml
+++ b/.github/workflows/check-ci.yml
@@ -33,6 +33,16 @@ jobs:
       - name: Run clippy
         run: cargo clippy --all-targets --verbose -- -D warnings
 
+      - name: Generate contract schemas
+        run: ./devtools/schema.sh
+
+      - name: Show schema changes
+        run: git status --porcelain
+
+      - name: Validate committed schemas
+        # should fail if any diffs are present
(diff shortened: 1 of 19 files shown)


keyleu:
can you separate the build from the schema check? Since most of CI depends on build but not on schemas we can have a separate action for checking the schemas that doesn't delay the rest

bekauz:
yeah I was thinking of that as well. feels like wasm optimizer steps can also sit in a different job as wasms are not relevant to regular unit testing. will just need to set up caching so that it doesn't rebuild stuff unnecessarily.

keyleu:
> yeah I was thinking of that as well. feels like wasm optimizer steps can also sit in a different job as wasms are not relevant to regular unit testing. will just need to set up caching so that it doesn't rebuild stuff unnecessarily.

they are used for test-tube tests

bekauz:
yeah sorry by "regular unit tests" i meant the non-integration ones

stiiifff (approved):

PR #104, opened 2024-11-19
Update assumptions.md

Art3miX:
This part is a little confusing. 
Sender is too vague, the code is not self explanatory here, because `authorization_contract` is referring to 2 different addresses based on which domain you on. We want to make sure it is clear.


keyleu:
This part of the docs refers to this:

```rust
            let authorized_sender = match config.processor_domain {
                ProcessorDomain::Main => config.authorization_contract,
                ProcessorDomain::External(polytone) => polytone.polytone_proxy_address.to_string(),
            };

            if info.sender.to_string() != authorized_sender {
                return Err(ContractError::Unauthorized(
                    UnauthorizedReason::NotAuthorizationModule {},
                ));
            }
```

PR #107, opened 2024-11-22
osmo cl e2e tests

### LINKED ISSUE #93: osmo e2e tests
with osmo gamm & CL services merged, we still need to cover their integration with an e2e test.

usually individual e2e tests for each type (gamm/CL) would be nice, but given the setup overhead of local-ic I think it makes sense to combine both of them into a single run that would involve both types (ie via splitter).

bekauz:
closes #93


contracts/authorization/src/tests/processor.rs
```diff
@@ -2727,10 +2725,7 @@ fn refund_and_burn_tokens_after_callback() {
             messages: vec![message1.clone()],
             ttl: None,
         }),
-        &[Coin {
-            denom: permission_token.clone(),
-            amount: "1".to_string(),
-        }],
+        &[Coin::new(Uint128::one(), permission_token.to_string())],
```
keyleu, line 2728: nice

contracts/libraries/osmosis-cl-withdrawer/src/contract.rs
```diff
@@ -142,16 +164,30 @@ fn handle_liquidity_withdrawal_reply(
...
+    // both coins cannot be zero because that would mean the position
+    // had no underlying liquidity to withdraw, so we skip the empty
+    // array check here and just fire the banksend
     let transfer_msg = BankMsg::Send {
         to_address: cfg.output_addr.to_string(),
-        amount: vec![
-            Coin::new(Uint128::from_str(&decoded_resp.amount0)?, pool.token0),
-            Coin::new(Uint128::from_str(&decoded_resp.amount1)?, pool.token1),
-        ],
+        amount: transfer_coins,
```
keyleu, line 190: Don't think for CL they are ordered btw, just looking at CL pools on osmosis and don't think they behave as GAMM. I also don't think it would matter as if it's not using proto, the bankmsg from wasm orders them, prob worth confirming it

bekauz: ah by ordering here I was referring to that we make sure that `amt_0` corresponds to `pool.token0` & same for `_1` to not mix up the denoms

keyleu: right

keyleu (approved):

PR #109, opened 2024-11-24
add pre_validate

Art3miX:
To mock the api used in contracts, we are using the `MockApi` type from the cosmwasm_std testing module, the api is only used to verify addresses, so we only need to provide the prefix and the MockApi should be working correctly.

The other big change is that we can't validate the config before we update the config with the addresses instead of ids of the accounts, so now instead of instantiating a library right after we change the config, we first loop over libraries and change the config for all of them, and pre_validate them, and then we loop over libraries again and instantiate them.
This is better because now before we instantiate any library, we first check them all.

To optimize even more, we might want to move the processor and authorization instantiation to after we validate the libraries, this way we will have a complete separation between "setup" step and "instantiate" step.


program-manager/src/library.rs
```diff
@@ -274,7 +274,7 @@ impl LibraryConfig {
         .map_err(LibraryError::SerdeJsonError)
     }
 
-    pub fn soft_validate_config(&self, api: &dyn cosmwasm_std::Api) -> LibraryResult<()> {
+    pub fn pre_validate_config(&self, api: &dyn cosmwasm_std::Api) -> LibraryResult<()> {
```
stiiifff, line 277: I think the Generic & Neutron IBC Transfer libs are missing in here.

program-manager/src/domain/cosmos_cw.rs
```diff
@@ -83,7 +83,7 @@ pub struct CosmosCosmwasmConnector {
     wallet: Wallet,
     code_ids: HashMap<String, u64>,
     chain_name: String,
-    prefix: String,
+    prefix: &'static str,
```
stiiifff, line 86: Why do you need a static str here ?

Art3miX: Fixed

program-manager/src/library.rs
Art3miX: Added

program-manager/src/domain/cosmos_cw.rs
```diff
@@ -133,6 +134,8 @@ impl CosmosCosmwasmConnector {
             code_ids: code_ids.clone(),
             chain_name: chain_info.name.clone(),
             prefix: chain_info.prefix.clone(),
+            api: cosmwasm_std::testing::MockApi::default()
+                .with_prefix(Box::leak(Box::new(chain_info.prefix.clone()))),
```
bekauz, line 138: I can't think of a way where this can go wrong but it'd be nice to add some comments about why `Box::leak` is fine here. Guess we could also have a chain prefix enum (with to/from string impl) which would make it static so we wouldn't need to deal with boxing like this?

bekauz (approved):

program-manager/src/domain/cosmos_cw.rs
Art3miX: Thanks @bekauz and @stiiifff 
We have a weird play with the connectors, because they are not cached for the entire program, because the manager is not a program, rather a library, so they are cached for the functions (init , update, migrate).
This was leaking the prefix on every time we call a function, not much, and can easily fixed with a server restart, but still, it is a leak.

I just copied the MockApi from cosmwas_std, and changed the api to accept String instead of the static str, which doesn't leak memory now.

Art3miX: also allowed us to remove our helpers for it and just use the mock api.

program-manager/src/mock_api.rs
stiiifff, line 1: It would be good to add a comment mentioning from which repo the code was copied, as well as which tag / commit. Also, add a license header to the file (from the original project).

PR #115, opened 2024-11-28
Token swap example program

stiiifff:


docs/src/examples/token_swap.md
```diff
@@ -1,13 +1,36 @@
...
+- **Party B Withdraw account**: the account from which Party B can withdraw Party A's tokens after the swap has successfully completed. Note: this can be a [Valence account](../components/accounts.md), but it could also be a regular chain account, or a smart contract.
+- **Party B Deposit account**:  a [Valence account](../components/accounts.md) which Party B will deposit their funds into, to be exchanged with Party A's funds.
+- **Splitter B**: an instance of the [Splitter library](../libraries/splitter.md) that will transfer Party B's tokens from its input account (i.e. the **Party B Deposit account**) to its output account (i.e. the **Party A Withdraw account**) upon execution of its `swap` function.
+- **Party A Deposit account**: the account from which Party A can withdraw Party B's tokens after the swap has successfully completed. Note: this can be a [Valence account](../components/accounts.md), but it could also be a regular chain account, or a smart contract.
```
bekauz, line 29:
```suggestion
- **Party A Withdraw account**: the account from which Party A can withdraw Party B's tokens after the swap has successfully completed. Note: this can be a [Valence account](../components/accounts.md), but it could also be a regular chain account, or a smart contract.
```

docs/src/examples/token_swap.md
```diff
@@ -1,13 +1,36 @@
...
+	subgraph Neutron
 	InA --> SSA --> OutB
 	InB --> SSB --> OutA
+	end
 ```
+
+The program is composed of the following components:
+
+- **Party A Deposit account**: a [Valence account](../components/accounts.md) which Party A will deposit their tokens into, to be exchanged with Party B's tokens.
+- **Splitter A**: an instance of the [Splitter library](../libraries/splitter.md) that will transfer Party A's tokens from its input account (i.e. the **Party A Deposit account**) to its output account (i.e. the **Party B Withdraw account**) upon execution of its `swap` function.
```
bekauz, line 25: guessing that by `swap` function you mean `split` here?

bekauz (commented):
left a few small comments. other than that, I wonder, couldn't we do this with a single reverse splitter?

keyleu:
Just one comment, maybe worth pointing out that the Deposit and Withdraw account of both parties can be the same account so they would only need 1 account

keyleu (approved):

PR #119, opened 2024-12-03
storage account v1

### LINKED ISSUE #118: storage account

bekauz:
closes #118 

this is the first version of a storage account meant to:
1. accept arbitrary data being posted
2. make it available via queries

this account will be used for #117  , so changes within this scope are to be expected.

opening this now to avoid a massive pr later down the line.

### DIFF
--- a/contracts/accounts/storage_account/.cargo/config.toml
+++ b/contracts/accounts/storage_account/.cargo/config.toml
@@ -1,0 +1,5 @@
+[alias]
+wasm = "build --release --lib --target wasm32-unknown-unknown"
+unit-test = "test --lib"
+schema = "run --bin schema"
+integration-test = "test --lib integration_tests"
--- a/contracts/accounts/storage_account/.cargo/config.toml
+++ b/contracts/accounts/storage_account/.cargo/config.toml
@@ -1,5 +1,3 @@
 [alias]
 wasm = "build --release --lib --target wasm32-unknown-unknown"
-unit-test = "test --lib"
 schema = "run --bin schema"
-integration-test = "test --lib integration_tests"
--- a/contracts/accounts/storage_account/Cargo.toml
+++ b/contracts/accounts/storage_account/Cargo.toml
@@ -1,0 +1,43 @@
+[package]
+name    = "valence-storage-account"
+version = "0.1.0"
+authors = ["Timewave Labs"]
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[lib]
+crate-type = ["cdylib", "rlib"]
+
+[features]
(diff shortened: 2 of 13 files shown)


contracts/accounts/storage_account/Cargo.toml
```diff
@@ -0,0 +1,43 @@
...
+[features]
+# use library feature to disable all instantiate/execute/query exports
+library = []
+
+[package.metadata.scripts]
+optimize = """docker run --rm -v "$(pwd)":/code \
+  --mount type=volume,source="$(basename "$(pwd)")_cache",target=/target \
+  --mount type=volume,source=registry_cache,target=/usr/local/cargo/registry \
+  cosmwasm/optimizer:0.16.0
+"""
```
keyleu, line 21: nit: can remove this

contracts/accounts/storage_account/.cargo/config.toml
```diff
@@ -0,0 +1,5 @@
+[alias]
+wasm = "build --release --lib --target wasm32-unknown-unknown"
+unit-test = "test --lib"
+schema = "run --bin schema"
+integration-test = "test --lib integration_tests"
```
keyleu, line 5: nit: can prob remove unit-test and integration-test

contracts/accounts/storage_account/README.md
keyleu, line 1: For some reason I thought we were going to make our current accounts be able to also store data on top of execution functions, but I guess separating them is also fine. Just prob worth to confirm this?

bekauz: main reason I separated it was to respect the existing _base_ account concept. I could see both types being merged into one, but at that point I think we should drop the notion of base accounts and just call it valence account which would encompass everything by itself? also `account-utils` package kinda loses it purpose at that point.

also, changing that would've required changes in both `account-utils` & `base-account` and I think those are frozen for the time being?

keyleu: sure, makes sense to me.

bekauz: if we do end up having different types of accounts I think it would make sense to revisit what we consider a base account as well. e.g., I'm not sure if every account should provide the `ExecuteSubmsgs` or even `ExecuteMsg` ? I thought about it for a while for this pr and ended up not including either because I couldn't think of a use case for it given the scope of a storage account (store binaries & make them available for querying).

keyleu:
Now that I think about it, probably worth updating the docs Account section as well including this new functionality

bekauz:
> Now that I think about it, probably worth updating the docs Account section as well including this new functionality

fair, I adjusted the existing accounts section to refer to *base account* instead of just an account.
added a section for the storage account, but left it with todo for now as this version of the account is just for testing / poc.

PR #122, opened 2024-12-12
EVM/Solidity processor part 1

keyleu:
First PR for our EVM support.

Includes: 
- CI for solidity contracts
- Simple processor that instantiates with a mailbox and 2 queues + simple testing.
- Queue implementation + testing.
- ProcessorMessage decoding (this is what will be sent from the authorization contract after passing through the encoder) with all the information that we need to construct the batch + testing.

The idea is to make incremental PRs in steps or it will get too hard to review.

### DIFF
--- a/solidity/.gitignore
+++ b/solidity/.gitignore
@@ -1,0 +1,11 @@
+# Compiler files
+cache/
+out/
+
+# Ignores development broadcast logs
+!/broadcast
+/broadcast/*/31337/
+/broadcast/**/dry-run/
+
+# Soldeer
+/dependencies
--- a/solidity/src/processor/Processor.sol
+++ b/solidity/src/processor/Processor.sol
@@ -1,0 +1,65 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.28;
+
+import {IMessageRecipient} from "hyperlane/interfaces/IMessageRecipient.sol";
+import {QueueMap} from "./libs/QueueMap.sol";
+
+contract Processor is IMessageRecipient {
+    // Use the library for the Queue type
+    using QueueMap for QueueMap.Queue;
+
+    // Authorization contract will be the sender of the messages on the main domain
+    // It's the hex representation of the address in the main domain
+    bytes32 public authorizationContract;
+
+    // Add mailbox address which will be the only address that can send messages to the processor
+    address public mailbox;
+
+    // Flag to check if the processor is paused
+    bool public paused;
+
+    // Declare the two queues
+    QueueMap.Queue private highPriorityQueue;
+    QueueMap.Queue private mediumPriorityQueue;
+
+    // Event declarations
+    event BatchAdded(string queueType, bytes data);
+    event BatchRemoved(string queueType, bytes data);
+    event MessageReceived(uint32 indexed origin, bytes32 indexed sender, bytes body);
+
+    // Custom errors
+    error UnauthorizedAccess();
+    error NotAuthorizationContract();
+    error InvalidAddress();
+    error ProcessorPaused();
+
+    constructor(bytes32 _authorizationContract, address _mailbox) {
+        // Check for zero addresses
+        if (_mailbox == address(0)) {
+            revert InvalidAddress();
+        }
+
+        // Set authorization contract and mailbox
+        authorizationContract = _authorizationContract;
+        mailbox = _mailbox;
+
+        // Initialize both queues with unique namespaces
+        highPriorityQueue = QueueMap.createQueue("HIGH_PRIORITY_QUEUE");
+        mediumPriorityQueue = QueueMap.createQueue("MEDIUM_PRIORITY_QUEUE");
+    }
+
+    // Implement the handle function
+    function handle(uint32 _origin, bytes32 _sender, bytes calldata _body) external payable override {
+        // Only mailbox can call this function
+        if (msg.sender != mailbox) {
+            revert UnauthorizedAccess();
+        }
+
+        // Check that the sender of the message is the authorization contract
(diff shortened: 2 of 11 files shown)


solidity/.gitignore
```diff
@@ -0,0 +1,11 @@
+# Compiler files
+cache/
+out/
+
+# Ignores development broadcast logs
+!/broadcast
+/broadcast/*/31337/
+/broadcast/**/dry-run/
```
bekauz, line 8: is this negation of broadcast `!/broadcast` needed if we later ignore those two paths?

solidity/src/processor/Processor.sol
```diff
@@ -0,0 +1,65 @@
...
+        authorizationContract = _authorizationContract;
+        mailbox = _mailbox;
+
+        // Initialize both queues with unique namespaces
+        highPriorityQueue = QueueMap.createQueue("HIGH_PRIORITY_QUEUE");
+        mediumPriorityQueue = QueueMap.createQueue("MEDIUM_PRIORITY_QUEUE");
+    }
+
+    // Implement the handle function
+    function handle(uint32 _origin, bytes32 _sender, bytes calldata _body) external payable override {
```
bekauz, line 52: should this be `payable`?

solidity/src/processor/Processor.sol
```diff
@@ -0,0 +1,65 @@
...
+        if (_mailbox == address(0)) {
+            revert InvalidAddress();
+        }
+
+        // Set authorization contract and mailbox
+        authorizationContract = _authorizationContract;
+        mailbox = _mailbox;
+
+        // Initialize both queues with unique namespaces
+        highPriorityQueue = QueueMap.createQueue("HIGH_PRIORITY_QUEUE");
```
bekauz, line 47: I think we should pass shorter namespace ids to save on gas costs for queue operations as we call `abi.encodePacked` on it, so shorter namespace will require less encoding?
```suggestion
        highPriorityQueue = QueueMap.createQueue("HPQ");
```

solidity/src/processor/libs/ProcessorMessage.sol
```diff
@@ -0,0 +1,367 @@
...
+        Indefinitely,
+        Amount
+    }
+
+    /// @notice Specifies how many times to retry
+    /// @param retryType Whether to retry indefinitely or a specific amount
+    /// @param amount Number of times to retry if type is Amount
+    struct RetryTimes {
+        RetryTimesType retryType;
+        uint64 amount;
```
bekauz, line 35: if RetryTimesType is `Indefinitely`, will amount be 0 and just be ignored?

solidity/src/processor/libs/ProcessorMessage.sol
```diff
@@ -0,0 +1,367 @@
...
+    /// @param message Raw message bytes
+    /// @return MessagesResult Parsed messages and next offset
+    function messages(bytes calldata message) internal pure returns (MessagesResult memory) {
+        uint64 count = numMessages(message);
+        bytes[] memory result = new bytes[](count);
+
+        uint256 offset = 26; // Start after header (1 + 8 + 1 + 8 + 8 bytes)
+
+        for (uint64 i = 0; i < count; i++) {
+            uint64 length = uint64(bytes8(message[offset:offset + 8]));
```
bekauz, line 154: should we validate that `offset:offset + 8` is a valid index in the message?

solidity/src/processor/libs/ProcessorMessage.sol
```diff
@@ -0,0 +1,367 @@
...
+        // Parse 20 bytes for address
+        address contractAddress = address(bytes20(message[offset:offset + 20]));
+        offset += 20;
+        func.contractAddress = contractAddress;
+
+        func.hasRetryLogic = message[offset] != 0;
+        offset += 1;
+
+        if (func.hasRetryLogic) {
+            (func.retryLogic, offset) = parseRetryLogic(message, offset);
```
bekauz, line 294: maybe we can add some comments here about how `parseRetryLogic` increments the offset? was a little confused before looking there

keyleu: Going to rework it, not worth adding comments for now as it wont be present in next PR

solidity/src/processor/libs/ProcessorMessage.sol
keyleu: Will be reworked, but in theory no validation would be needed because we are doing the encoding as well and would ensure it would be a valid index

solidity/src/processor/libs/ProcessorMessage.sol
keyleu: This will all be reworked but yes, the idea was to send something there and ignore it.

solidity/src/processor/Processor.sol
keyleu: I'm overriding it directly from Hyperlane, where it is declared as:
```solidity
interface IMessageRecipient {
    function handle(
        uint32 _origin,
        bytes32 _sender,
        bytes calldata _message
    ) external payable;
}
```

So answer yes. Reason I can think of is you modify the Mailbox contract to include funds there so they don't make it non payable.

solidity/.gitignore
keyleu: removed all this for now as we don't have the subdirectory. These lines were added automatically by forge. We're going to do integration tests with local-ic so doubt we are going have them later.

solidity/src/processor/Processor.sol
bekauz: got it

bekauz (approved):

PR #124, opened 2024-12-13
Audit: validate fields during ibc-transfer config update

keyleu:
Add validation of channel_id and ibc transfer timeout when updating ibc transfer library configs


bekauz (approved):
lgtm

PR #127, opened 2024-12-19
Audit: fix astroport ratio calculation

keyleu:
fix the ratio calculation that was being done wrong in the double sided liquidity when we didn't have enough tokens of 1 asset as per audit feedback.

### DIFF
--- a/contracts/libraries/astroport-lper/src/contract.rs
+++ b/contracts/libraries/astroport-lper/src/contract.rs
@@ -208,7 +208,7 @@ mod functions {
             Ok((required_asset1_amount.u128(), balance2))
         } else {
             // We can't provide all asset2 tokens so we need to determine how many we can provide according to our available asset1
-            let ratio = cosmwasm_std::Decimal::from_ratio(pool_asset1_balance, pool_asset2_balance);
+            let ratio = cosmwasm_std::Decimal::from_ratio(pool_asset2_balance, pool_asset1_balance);
 
             Ok((
                 balance1,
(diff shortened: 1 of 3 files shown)


Art3miX (approved):

contracts/libraries/astroport-lper/src/contract.rs
```diff
@@ -208,7 +208,7 @@ mod functions {
             Ok((required_asset1_amount.u128(), balance2))
         } else {
             // We can't provide all asset2 tokens so we need to determine how many we can provide according to our available asset1
-            let ratio = cosmwasm_std::Decimal::from_ratio(pool_asset1_balance, pool_asset2_balance);
+            let ratio = cosmwasm_std::Decimal::from_ratio(pool_asset2_balance, pool_asset1_balance);
```
bekauz, line 211: hmm out of curiosity I checked the cosmwasm 2.1.4 `Decimal::from_ratio` impl and it just silently panics:
```rust
    /// Returns the ratio (numerator / denominator) as a Decimal
    pub fn from_ratio(numerator: impl Into<Uint128>, denominator: impl Into<Uint128>) -> Self {
        match Decimal::checked_from_ratio(numerator, denominator) {
            Ok(value) => value,
            Err(CheckedFromRatioError::DivideByZero) => {
                panic!("Denominator must not be zero")
            }
            Err(CheckedFromRatioError::Overflow) => panic!("Multiplication overflow"),
        }
    }
```

should we just do `checked_from_ratio` instead? will need to map the `CheckedFromRatioError` to `LibraryError` but I think it may be worth it as we get those pool_asset2_balance/pool_asset1_balance with `querier.query_balance` which can actually return 0 (unlike query_all_bals which would emit it)?

bekauz (approved):

PR #130, opened 2024-12-23
Audit: burn token when operation removed by owner

keyleu:
When a token was used to execute an operation but that operation was forcefully removed by the owner, the token will be burned instead of sent back. If the owner wants to mint a new one then he can do that with an extra mint operation.


contracts/authorization/src/tests/processor.rs
```diff
@@ -2830,6 +2830,153 @@ fn refund_and_burn_tokens_after_callback() {
...
+            label: "permissioned-with-limit".to_string(),
+            messages: vec![message1.clone()],
+            ttl: None,
+        }),
+        &[Coin::new(Uint128::one(), permission_token.to_string())],
+        &setup.user_accounts[0],
+    )
+    .unwrap();
+
+    // Let's balance of the user to verify that he doesn't have any tokens left
```
bekauz, line 2909:
```suggestion
    // check the balance of the user to verify that they don't have any tokens left
```

bekauz (commented):
lgtm. can we just add a test case for `PartiallyExecuted` execution result for the future to protect from some accidental changes?

contracts/authorization/src/tests/processor.rs
keyleu: fixed!

keyleu:
> lgtm. can we just add a test case for `PartiallyExecuted` execution result for the future to protect from some accidental changes?

Good catch, thought we had one. Added a check that tokens are also burned with PartiallyExecuted

PR #139, opened 2025-01-08
Update authorization_instantiation.md

hxrts:
small readme formatting update


keyleu (approved):

keyleu:
just noticed right below this part there is a mention of "service" instead of library, "instantiating the services". If you can change that one too before merging would be nice

PR #143, opened 2025-01-10
initial middleware docs

### LINKED ISSUE #140: add middleware to docs
middleware type-registries and broker already contain brief readme descriptions.
elaborate on them and add their dedicated entries to the docs.

bekauz:
closes #140 

introduces the initial version of the middleware documentation that includes descriptions of:

- Valence Types
- Middleware Brokers
- Middleware Type Registries

this work is by no means complete; instead, it is meant to serve as a basis for more productive discussions of subjects that depend on it (such as  #117).


docs/src/middleware/broker.md
```diff
@@ -0,0 +1,72 @@
...
+
+Among other things, Middleware brokers should enable such programs to gracefully
+recover into a synchronized state that can continue operating in a correct manner.
+
+## Broker Lifecycle
+
+Brokers are singleton components that are instantiated before the program start
+time.
+
+Valence Programs refer to their brokers of choice by their respective addresses.
```
uditvira, line 43: Which component of Valence Programs should refer to the broker?

bekauz: any component is free to do that.

if it's a library, then the logic needs to be encoded in the actual code (querying the broker and performing actions based on the response).

if it's a processor, then an assumption is made that it is the one initiating the queries and evaluating the responses (in some kind of stack-resembling fashion). if the evaluation checks out, some action may be performed.

keyleu (commented):
Makes sense to me because I now understand how the ICQ relayer, but I can imagine this can be confusing for program creators. I'd personally add an example of how this can be used: like a small diagram of how this all fits in a real program (i.e. what actions I need to take to use this in my program).

One thing that could also be useful to me is a summary of the steps I need to take to add a new type and add that to my query as it might not be that straightforward. For example, for the KV key there is a prefix and an id, which we take directly from the osmosis chain, but someone might not know this.

keyleu (approved):

PR #147, opened 2025-01-14
liquidity withdrawer pool ratio checks

### LINKED ISSUE #145: adapt liquidity withdrawers to validate pool ratio
## Issue

Between the moment when liquidity withdrawal message is queued and its actual execution, liquidity in the pool (and its ratio) is subject to fluctuations due to other user trades.

This applies to both astroport- and osmosis-liquidity withdrawers.

## Solution

The expected coin output should be taken in as a parameter, similarly to how it's done in the liquidity providers.
During the withdrawal function execution, it should validate the expected coin output against the user specified parameter and only proceed with the withdrawal in case the expected outcome is inline with user expectations.

bekauz:
closes #145 

both `valence-osmosis-gamm-withdrawer` and `valence-astroport-withdrawer` now take in `expected_pool_ratio_range: Option<DecimalRange>` as an argument for liquidity withdrawal. this range gets validated against the current pool ratio: if the current pool ratio is not contained in the specified `DecimalRange` (if any), function returns an error. if `DecimalRange` is not specified, pool ratio validation is skipped.

in order to enable that price validation, both astroport & osmosis withdrawer `LiquidityWithdrawerConfig`s now take in an additional parameter - `AssetData`, specifying the denoms in the pool. this pattern follows the same logic as liquidity providers.

because of how much more similar liquidity providers and withdrawers are now looking, a bunch of stuff was moved into utils and normalized across osmosis and astroport counterparts:
- `AssetData`
- `DecimalRange`

### summary

- same `DecimalRange` and `AssetData` structs are now used by `valence-osmosis-gamm-withdrawer`, `valence-osmosis-gamm-lper`, `valence-astroport-withdrawer`, and `valence-astroport-lper`
- `valence-astroport-withdrawer` and `valence-osmosis-gamm-withdrawer` perform decimal range validation on withdrawal
- for astroport components, the following have been moved to utils for reuse: `AssetTrait`, `PoolType`, fns for query helpers, `get_pool_asset_amounts` 

#### out of scope changes

I noticed that astroport lper had not unit tested the decimal range validation on execution, so some unit tests were added on that end.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5073,6 +5073,7 @@ dependencies = [
  "cosmwasm-schema 2.1.4",
  "cosmwasm-std 2.1.4",
  "neutron-test-tube",
+ "valence-library-utils",
 ]
 
 [[package]]
--- a/contracts/libraries/astroport-lper/src/astroport_cw20.rs
+++ b/contracts/libraries/astroport-lper/src/astroport_cw20.rs
@@ -1,18 +1,12 @@
-use crate::msg::{Config, PoolType};
+use crate::msg::Config;
 
 use cosmwasm_std::{coin, CosmosMsg, WasmMsg};
 use cosmwasm_std::{to_json_binary, DepsMut, Uint128};
 use valence_astroport_utils::astroport_cw20_lp_token::{
-    Asset, AssetInfo, ExecuteMsg, PairType, PoolQueryMsg, PoolResponse, SimulationResponse,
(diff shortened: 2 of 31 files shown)


bekauz:
@keyleu if you can give a closer look to the whole astroport versioning aspect, that'd be great. I moved a bunch of astroport logic to utils that is now reused across both withdrawer & provider, so just want to make sure nothing went wrong there

keyleu (approved):
Looks good to me, also very nice clean up overall, the astroport libraries needed this, thx for that

PR #148, opened 2025-01-14
Hyperlane set up with local-ic

keyleu:
This PR adds all the groundwork for testing with Hyperlane and local-ic.

- Adds all the helper functions for deploy and instantiate all hyperlane contracts on CW and EVM.
- Does all the docker setup: creates a new network, adds the containers to the network, updates the relayer config with the new contracts that were just deployed/instantiated and finally spins up the relayer between the containers. It's all set up in a way that it can be easily rerun and takes care of everything: stops and removes the previous relayer, does the cleanup and runs a new fresh relayer.
- The EthClient was moved to localic-utils because it makes more sense there but it's just a blocker on the async operations.
- Provided an example of how it works (this example will later be modified using it with the authorization contract instead of interacting with the mailbox directly).
- Updated README with instructions, only additional thing that needs to be done is run the ethereum testnet node with anvil with the command described.
- The private key used to sign messages with the relayer is the one of the keys anvil provides (the last one). For callbacks we'll need to fund the neutron address that private key gives but I'll add that when it's actually necessary (just a bank transfer to the address).


local-interchaintest/examples/ethereum_test.rs
```diff
@@ -0,0 +1,98 @@
...
+        .build()?;
+
+    // Upload all Hyperlane contracts to Neutron
+    let neutron_hyperlane_contracts = set_up_cw_hyperlane_contracts(&mut test_ctx)?;
+    // Deploy all Hyperlane contracts on Ethereum
+    let eth_hyperlane_contracts = set_up_eth_hyperlane_contracts(&eth, 1)?;
+
+    set_up_hyperlane(
+        "hyperlane-net",
+        vec!["localneutron-1-val-0-neutronic", "anvil"],
```
bekauz, line 37: can probably get `localneutron-1-val-0-neutronic` from `test_ctx`, no?

local-interchaintest/examples/ethereum_test.rs
```diff
@@ -0,0 +1,98 @@
...
+    let address_hex = test_recipient_address
+        .to_string()
+        .trim_start_matches("0x")
+        .to_string();
+    // Pad to 32 bytes (64 hex characters) because mailboxes expect 32 bytes addresses with leading zeros
+    let padded_recipient = format!("{:0>64}", address_hex);
+    let msg_body = HexBinary::from_hex(&hex::encode("Hello my friend!"))?;
+
+    let dispatch_msg = hpl_interface::core::mailbox::ExecuteMsg::Dispatch(DispatchMsg {
+        dest_domain: 1,
```
bekauz, line 69: is this `1` referring to some hyperlane relayer state property?

local-interchaintest/README.md
```diff
@@ -18,6 +18,14 @@ local-ic start neutron_juno --api-port 42069
 
 This will start a local environment with a Gaia chain, a Neutron (using ICS) chain and a Juno chain. The `--api-port` will expose the API on port 42069, we are using this port in our local-ic-utils crate so let's use the same to reuse some of the utils there.
 
+To spin up a local Ethereum testnet node, we use Anvil. If you want to run the tests that involve Ethereum, you need to start the Anvil node beforehand. You can do this by running the following command:
```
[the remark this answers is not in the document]
keyleu: added to spin it up from code and removed from readme

local-interchaintest/examples/ethereum_test.rs
keyleu: couldn't find it in test_ctx, in any case it's always the same so rather just have it like this for simplicity

local-interchaintest/examples/ethereum_test.rs
keyleu: yeah I'm telling to which domain the relayer has to relay too. Ethereum is domain id 1

keyleu: for reference https://hyp-v3-docs-er9k07ozr-abacus-works.vercel.app/docs/reference/domains

local-interchaintest/README.md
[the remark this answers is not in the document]
bekauz: awesome. I think "not spin up again if it exists" makes sense. for what its worth, I think the ICQ relayer tests do "remove if exists; spin up" which is basically the same.

bekauz (approved):
just did a second pass and the whole setup looks great, awesome work!

only thing that I can imagine potentially going wrong is something with github runner environment not playing nice with the networking config (localhost rpc ports and all). guess we could quickly check that by changing the destination branch of this pr to be `main` to trigger the jobs, but nevertheless that fix would be minimal. lgtm!

PR #149, opened 2025-01-15
neutron ic querier docs

bekauz:
work towards #117 
builds upon #143 (this should be merged after middleware docs)

this pr introduces the initial version of the Neutron IC Querier docs.
this work is not complete - multiple sections in the pr propose ideas and potential directions rather than outline some existing outcome.

in addition to the IC Querier section, it will also bring in the storage account documentation.


docs/src/libraries/neutron_ic_querier.md
```diff
@@ -0,0 +1,264 @@
...
+it available for further processing, interpretation, or other functions.
+
+## Library Lifecycle
+
+With the baseline functionality in mind, there are multiple design decisions to be made
+that will shape the overall lifecycle of this library.
+
+This library may take on one of two possible routes with respect to how long it is considered
+active:
+1. single-use, instantiated to perform a particular query and complete
```
uditvira, line 200: Single-use makes sense to me purely to stay consistent with other libraries. But single use only means that it is associated with a fixed set of output (storage) accounts. You could consider it making multiple queries that all get written to that same account.

docs/src/libraries/neutron_ic_querier.md
```diff
@@ -0,0 +1,264 @@
...
+2. multi-use, instantiated to perform multiple domain queries
+
+For single-use design, one of the main questions is at what point does the query
+actually get registered.
+
+One approach for the single-use approach could be to register the query as soon as
```
uditvira, line 206: Yes, we should allow the Library to be instantiated with a pre-configured queries.

docs/src/libraries/neutron_ic_querier.md
```diff
@@ -0,0 +1,264 @@
...
+funds, Neutron IC Querier has a notion of output account for writing some data.
+
+For example, consider a situation where this library had queried the balance of some
+remote account, parsed the response into a Valence Balance type, and wrote that resulting
+object into its associated output account. That output account may be the input account
+of some other library, which will attempt to perform its function based on the content
+written to its input account. This may involve something along the lines of:
+`if balance > 0, do x; otherwise, do y;`.
+
+It is a little less obvious as to what is the *input account* of Neutron IC Querier.
```
uditvira, line 243: There is only one account associated with the Library and it's the account that the Library has permissions to write to the storage to. This breaks our previous conventions around Input/Output accounts. 

Previously, until now, Input accounts were accounts funds were take from and output accounts were accounts where funds were sent to. Or in other words, the Library had execute permissions on Input Accounts. The library did not have execute permissions around Output accounts.

In this case, we want the Library to have write permissions to the storage of an account.

docs/src/libraries/neutron_ic_querier.md
```diff
@@ -0,0 +1,264 @@
...
+For one, the input account could be standard type account that deals with token transfers.
+Prior to registering the interchain query, input account could be expected to receive
+the query deposit (in `untrn`) meant to cover the query registration costs.
```
uditvira, line 246: This is not what we do in the ntrn ibc transfer library which also requires a fee. That is the closest analogous library

uditvira (commented):
Let's discuss these in our call tomorrow.

docs/src/libraries/neutron_ic_querier.md
```diff
@@ -0,0 +1,264 @@
...
+
+    /// **value** is a value which is stored under the **key** in the storage on remote chain
+    pub value: Binary,
+}
+```
+
+## Query lifecycle
+
+After `RegisterInterchainQuery` message is submitted, `interchainqueries` module will deduct
+the query registration fee from the caller.
```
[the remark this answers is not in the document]
bekauz: I see your point here. I left it a little ambiguous on purpose because I didn't want to make assumptions on who the actual caller may be in the final design - account or library. Regardless, if it is the library, what do you think of:
> from the caller (which is this Library).

Obviously I'm far from an expert in grammar but the structure "After `RegisterInterchainQuery` message is submitted, <...>, this Library." made me a little confused on the first pass. Could be skill issue on my end.

docs/src/libraries/neutron_ic_querier.md
```diff
@@ -0,0 +1,264 @@
...
+For a given KV-query to be performed, the underlying type registry must implement `IcqIntegration` trait
+which thefore provides the following adapters:
+1. `get_kv_key`, enabling the ability to get the correct `KVKey` for query registration
+2. `decode_and_reconstruct`, allowing to reconstruct the interchain query result
+
+Read more about the given type ICQ integration in the [type registry documentation page](./../middleware/type_registry.md#neutron-interchain-query-integration).
+
+### Query registration fee
+
+Neutron `interchainqueries` module is configured to escrow a fee (denominated in `untrn`) in order to register a query.
```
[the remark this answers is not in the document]
bekauz: yeah true. what do you think if we introduced a markdown linter to break lines everywhere consistently? we could run it along with the documentation build job for consistency

bekauz:
these docs are to become slightly out of date once #158 is merged. any unaddressed comments above & updates to the new design will be done in a follow up pr.

uditvira (approved):

PR #150, opened 2025-01-16
Eth vault

Art3miX:
Very basic implementation of the vault on Ethereum, this is a starting point to build on top of and add visibility for others.

### TODO:

- Override all functions of the interface to match with our custom logic of the vault
- Add custom entry points needed to support our design.
- Write tests that covers the core logic and our custom logic


solidity/src/libraries/ValenceVault.sol
```diff
@@ -0,0 +1,64 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.28;
+
+import {SafeERC20, IERC20Metadata, ERC20, IERC20, IERC4626, ERC4626} from "@openzeppelin-contracts/token/ERC20/extensions/ERC4626.sol";
+import {Math} from "@openzeppelin-contracts/utils/math/Math.sol";
```
keyleu, line 5: some of them not being used, not sure if leftovers

solidity/src/libraries/ValenceVault.sol
```diff
@@ -0,0 +1,64 @@
...
+import {BaseAccount} from "../accounts/BaseAccount.sol";
+
+contract ValenceVault is Library, ERC4626 {
+    struct VaultConfig {
+        BaseAccount DepositAccount;
+        BaseAccount WithdrawAccount;
+    }
+
+    VaultConfig public config;
+    address public strategist;
```
keyleu, line 16: should strategist be part of the VaultConfig too?

keyleu (approved):

solidity/src/libraries/ValenceVault.sol
Art3miX: Good point, moved

PR #155, opened 2025-01-19
eth vault

Art3miX:
Overridden functions to match our custom logic.
Modified `MaxDeposit` and `MaxMint` to add a total deposit cap.
Modified deposit to send funds to the `depositAccount` and not to the vault contract as default.
Changed the `_convertToAssets` and `_convertToShares` to reflect our custom conversions.
Added test for all of it.


solidity/test/libraries/Vault.t.sol
```diff
@@ -64,26 +73,199 @@ contract VaultTest is Test {
...
+
+        vault.deposit(depositAmount, user);
+
+        assertEq(token.balanceOf(address(depositAccount)), depositAmount);
+        assertEq(token.balanceOf(user), preBalance - depositAmount);
+        assertEq(vault.balanceOf(user), depositAmount);
+        assertEq(vault.totalSupply(), depositAmount);
+
+        vm.stopPrank();
+    }
```
keyleu, line 153: nit: I'd add a test for the convertToShares and convertToAssets after deposit to check that the rounding is working as expected since in the previous tests it's always multiplying by 1

keyleu (approved):

solidity/test/libraries/Vault.t.sol
Art3miX: Deposits doesn't change the rate, so it shouldn't change the rounding, but update does, I will do more tests regarding updates, I think already have those in the next PR.

PR #154, opened 2025-01-17
storage account rework

bekauz:
work towards #153 

# Changes

Reworks the storage account contract to store `ValenceType` variants instead of cosmwasm binaries.

Along with that, the api naming is updated:
- for execution, instead of `PostBlob` the storage variant is now called `StoreValenceType`
- for querying, instead of `Blob` the query is now `QueryValenceType`

PR #158, opened 2025-01-22
neutron ic querier follow-up

### LINKED ISSUE #153: neutron IC querier follow-up
Following the initial ic querier poc and subsequent specifications, the following changes should be applied.

# Neutron IC Querier changes

## Accounts

All that the IC Querier cares about with respect to accounts is which account it is permitted to write the results into.
The library should therefore have one account associated with it, and it should follow the same rules that usual 
account authorizations do.

## Lifecycle

The library should be instantiated with all desired query configurations.
Multiple queries can be configured on instantiation, as long as they are all going to be written to the same account.

Configured queries should be registered (triggered) by an execute method.
Registered queries should be deregistered by another execute method. 
 
# Storage Account changes

Storage accounts should no longer store arbitrary data blobs (binaries).
Instead, they should be dealing exclusively with Valence Types.

bekauz:
closes #153 

# Summary

Updates the previous neutron IC querier iteration to one that functions as a library and is compatible with Valence programs.

## Setup

Library is configured with the following parameters:

- `storage_account`, pointing to the target storage account where ICQ results are to be written
- `querier_config`, defining the middleware broker and the connection id of the target domain (applies to all defined queries)
- `query_definitions`, where target queries are configured as a map where the key is an arbitrary string (id) and the value is the actual query configuration (type url, update period, query parameters and an optional type registry version)

## Functions

Library provides two functions:

- `RegisterKvQuery { target_query: String }`, which takes in a configured query identifier and submits the query registration request to `interchainqueries` module
-  `DeregisterKvQuery { target_query: String }`, which takes in a configured query identifier and submits the query deregistration request and transfers the received escrow fee to the caller

Part of the functionality that deals with processing query results is hidden from the user. Instead, result processing is triggered by the `sudo` callback initiated by ICQ relayers. With that, the submitted query id is queried and processed into a `ValenceType` which then gets posted into the associated storage account. One thing to note here is that storage accounts (see #154) expect the types being posted to be associated with a storage key of type `String`. For that reason we reuse the same `target_query: String` identifier that we use in `RegisterKvQuery` call which is the key value from `query_definitions` configured on library setup.


keyleu (commented):
Implementation lgtm, we've already gone through some parts on calls as well. If all tests pass feel free to merge (remember to regenerate schema if not done)

bekauz:
@keyleu thanks for giving it a look! I slightly changed the deregistration logic just now - it no longer takes in the neutron query id and instead takes in the assigned query id from the library config. would appreciate if you could check that part again.

tests do pass and I added more explicit checks, but it's not yet enabled on the CI because of the ICQ relayer operating as a sidecar. i'll make an issue to integrate that.

Issue #151, opened 2025-01-16
[WIP] Eth Vault progress tracking


Art3miX:
This is an issue to track the eth vault progress.

#150 is the starting PR with the most basic implementation of the vault, all our custom logic is missing, and will added in steps to ensure fast  and stable progress toward the final result.

This will help us progress faster in implementing new designs, and will also enable use of tools (like AI) in the development stage.

## Related PRs
* #150 
* #155 
* #156 
* #161 
* #182 

## General goal

Finish the vault implementation by using a well defined steps from start to finish.

## Instructions

You are solidity expert and know everything about blockchains.
Think step by step and also follow instructions closely. don't over explain only if asked.
Always provide full, documented code as artefacts. 
Disagree with me if you believe a particular solution is not viable enough. Ask me to change direction if required.
Ask me any questions so you can give me the best possible response.
Only provide 1 artifact per file.

## Context

We have the figma design for the full implementation of the program, this includes the vault on Ethereum requirements.

We will be following the ERC4626 standard for the vault: https://ethereum.org/en/developers/docs/standards/tokens/erc-4626/

We will be using openzeppelin ERC4626 abstract contract to implement our vault: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v5.1.0/contracts/token/ERC20/extensions/ERC4626.sol#L135

We have the docs that layout the general design in text, but is missing specific implementation details: https://github.com/timewave-computer/valence-protocol/blob/main/docs/src/examples/crosschain_vaults.md

#### Strategist 

This is a trusted actor that can call some functions on the vault specifically.

#### Redemption rate

This can only be updated by the strategist every EPOCH.
This is the rate at which shares can be redeemed into assets, `TotalAssets() / TotalSupply()`

### Override standard functions to implement our custom logic

While the openzeppelin abstract contract gives us a base implementation of a vault, our custom logic require us to modify most of functions.

#### 1. _convertToAssets(uint256 shares, Math.Rounding rounding)

`shares * RedemptionRate`

#### 2. _convertToShares(uint256 assets, Math.Rounding rounding)

`assets / RedemptionRate`

#### 3. totalAssets()

Our total assets count is `TotalSupply() * RedemptionRate` 
Given the above helpers, we can use `_convertToAssets(TotalSupply(), Math.Rounding.Floor)`

#### 4. _deposit(address caller, address receiver, uint256 assets, uint256 shares)

[...]

PR #159, opened 2025-01-22
EVM integration into authorization contract

keyleu:
PR that includes the final steps for the EVM integration

Includes:

- Small refactoring to authorization creation API to be more accurate when we are adding an external domain: we can only add cosmwasm connectors to cosmwasm execution environment etc.
- Integration with the encoder and Hyperlane.
- Overall refactor of authorization contract without modifying the previous logic to separate new logic in a cleaner way.
- Extra validations during authorization creation to not allow creating authorizations that allow messages for the wrong execution environment or to provide invalid param restrictions for specific messages.
- Tests for all EVM scenarios.

Activated CI on dev branch as well to verify that everything is still working as expected after the changes we've been adding.


contracts/authorization/src/domain.rs
```diff
@@ -16,8 +21,8 @@ use crate::{
 pub fn add_domain(
     deps: DepsMut,
     callback_receiver: String,
-    domain: &ExternalDomainInfo,
-) -> Result<CosmosMsg, ContractError> {
+    domain: ExternalDomainInfo,
+) -> Result<Option<CosmosMsg>, ContractError> {
```
bekauz, line 25: something about the function definition here feels a little off, `add_domain` returns an optional cosmos message.

seems like currently it does the following:
- validates & saves the domain being added
- returns an optional cosmos msg

what if we split the two away? so that this fn would just validate & save, and the calling fn (`add_external_domains` in authorizations contract) would take care of building the message if it's relevant

contracts/authorization/src/domain.rs
```diff
@@ -72,45 +83,97 @@ pub fn get_domain(authorization: &Authorization) -> Result<Domain, ContractError
...
+
+/// Formats an address for Hyperlane by removing the "0x" prefix and padding it to 32 bytes
+pub fn format_address_for_hyperlane(address: String) -> Result<HexBinary, ContractError> {
+    // Remove "0x" prefix if present
+    let address_hex = address.trim_start_matches("0x").to_string().to_lowercase();
+    // Pad to 32 bytes (64 hex characters) because mailboxes expect 32 bytes addresses with leading zeros
+    let padded_address = format!("{:0>64}", address_hex);
+    // Convert to HexBinary which is what Hyperlane expects
+    Ok(HexBinary::from_hex(&padded_address)?)
+}
```
bekauz, line 179: feels like we could move these to hyperlane utils?

contracts/authorization/src/domain.rs
```diff
@@ -72,45 +83,97 @@ pub fn get_domain(authorization: &Authorization) -> Result<Domain, ContractError
...
+                contract_addr: processor,
+                msg: execute_msg,
+                funds: vec![],
+            })],
+            callback: callback_request,
+            timeout_seconds: Uint64::from(polytone_note.timeout_seconds),
+        })?,
+        funds: vec![],
+    }))
+}
```
bekauz, line 150: can we move this to polytone utils?

contracts/processor/src/contract.rs
```diff
@@ -501,7 +501,8 @@ fn execute_atomic(
             UnauthorizedReason::NotProcessor {},
         ));
     }
-    let messages: Vec<CosmosMsg> = batch.into();
+
+    let messages = Into::<Result<Vec<CosmosMsg>, StdError>>::into(batch)?;
```
bekauz, line 505:
```suggestion
    let messages = Into::<StdResult<Vec<CosmosMsg>>>::into(batch)?;
```

bekauz (commented):
done with the first pass, will revisit again during the week

packages/authorization-utils/src/domain.rs
```diff
@@ -11,54 +11,91 @@ pub enum Domain {
...
+        state: PolytoneProxyState,
+    ) -> Result<(), &'static str> {
+        match &mut self.execution_environment {
+            ExecutionEnvironment::Cosmwasm(CosmwasmBridge::Polytone(polytone_info)) => {
+                polytone_info.polytone_note.state = state;
+                Ok(())
             }
+            ExecutionEnvironment::Evm(_, _) => {
+                Err("EVM domain does not have a polytone proxy state")
+            }
```
Art3miX, line 60: Why not return an error here?

packages/authorization-utils/src/msg.rs
```diff
@@ -22,67 +25,137 @@ pub struct InstantiateMsg {
...
+impl ExternalDomainInfo {
+    pub fn into_external_domain(self, api: &dyn Api) -> StdResult<ExternalDomain> {
+        Ok(ExternalDomain {
+            name: self.name,
+            execution_environment: self.execution_environment.into_execution_environment(api)?,
+            processor: self.processor,
+        })
+    }
+}
+
```
Art3miX, line 158: Why is it split from the 2nd implementation? I would suggest keeping all those functions together.

contracts/authorization/src/contract.rs
```diff
@@ -953,6 +1085,82 @@ fn process_polytone_callback(
...
+    _env: Env,
+    info: MessageInfo,
+    handle_msg: HandleMsg,
+) -> Result<Response, ContractError> {
+    // We need to check that the callback comes from a registered processor address in the external domains and that the domain ID matches
+    // and obtain the encoder to decode the callback
+    let encoder = EXTERNAL_DOMAINS
+        .range(deps.storage, None, None, Order::Ascending)
+        .filter_map(Result::ok)
+        .find_map(|(_, external_domain)| {
```
Art3miX, line 1099: Do we really need the range here? seems like we should also keep record where the address is the key? 
Write once seems to be much better solution then loop on each callback.

contracts/authorization/src/contract.rs
```diff
@@ -953,6 +1085,82 @@ fn process_polytone_callback(
...
+
+                    match format_address_for_hyperlane(external_domain.processor.clone()) {
+                        Ok(formatted_address) => {
+                            if formatted_address == handle_msg.sender {
+                                Some(encoder.clone())
+                            } else {
+                                None
+                            }
+                        }
+                        Err(_) => None,
```
Art3miX, line 1126: Is this enough? 
I might be missing something, but it seems like the only validation that is not forgeable by an attacker is the mailbox address, the first one.
An attacker can format any message he wants in the `handle_msg` no?

contracts/authorization/src/contract.rs
```diff
@@ -953,6 +1085,82 @@ fn process_polytone_callback(
...
+    // Deserialize the callback to obtain the execution ID and the result
+    let InternalAuthorizationMsg::ProcessorCallback {
+        execution_id,
+        execution_result,
+    } = from_json(&wrapped_callback)?;
+
+    // Update the information
+    let mut callback_info = PROCESSOR_CALLBACKS.load(deps.storage, execution_id)?;
+    callback_info.execution_result = execution_result;
+    PROCESSOR_CALLBACKS.save(deps.storage, execution_id, &callback_info)?;
```
Art3miX, line 1157: I would use the update here, just a minor nit

Art3miX (commented):
Some nits and questions, overall looks good, even tho a bit long, but still manageable.

contracts/authorization/src/domain.rs
keyleu: good points, just separated it to not make them dependent on the other

packages/authorization-utils/src/domain.rs
bekauz, line 1: seeing the matches on `self.execution_environment` here makes me feel like it could be cleaner if these calls were a trait impl on `ExecutionEnvironment` enum itself

packages/bridging-utils/Cargo.toml
```diff
@@ -1,9 +1,9 @@
 [package]
-name        = "valence-polytone-utils"
+name        = "valence-bridging-utils"
```
bekauz, line 2: how about `valence-gmp-utils`? polytone doesn't deal with bridging assets

solidity/src/libraries/ValenceVault.sol
bekauz, line 1: on a high level this makes sense to me, although having a readme would be very helpful.

one high level question I have is: how are pending deposits treated during share issuance? for example, if funds are sitting in the `depositAccount` and someone wants to do a new deposit, are those pending funds taken into account?

packages/authorization-utils/src/domain.rs
Art3miX: To make the question clearer, why not return a `StdError` or custom error here?

solidity/src/libraries/ValenceVault.sol
Art3miX: I think the reason its in this PR is because of formatting, should ignore it here and read #151 for the vault thing.

contracts/authorization/src/domain.rs
keyleu: moved

contracts/processor/src/contract.rs
keyleu: changed!

contracts/authorization/src/authorization.rs
```diff
@@ -18,12 +19,25 @@ use crate::{
 };
 
 pub trait Validate {
-    fn validate(&self, store: &dyn Storage) -> Result<(), ContractError>;
+    fn validate(&self, store: &dyn Storage, querier: QuerierWrapper) -> Result<(), ContractError>;
```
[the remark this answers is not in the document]
keyleu: Just for code cleanliness, this groups everything related to validations

contracts/authorization/src/authorization.rs
```diff
@@ -80,37 +92,164 @@ impl Validate for Authorization {
...
+
+    /// For SolidityCall messages:
+    /// - Verifies library exists in encoder by querying the encoder broker
+    fn validate_solidity_library<T: Function>(
+        &self,
+        func: &T,
+        querier: &QuerierWrapper,
+    ) -> Result<(), ContractError> {
+        // If it's a solidity call, the library must be a valid library on the encoder
+        if let MessageType::SolidityCall(encoder, library_name) =
```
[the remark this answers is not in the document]
keyleu: Changed everywhere.

packages/authorization-utils/src/msg.rs
keyleu: wdym split from the 2nd implementation? I have everything here. Just separating transformations to keep things a bit more manageable

contracts/authorization/src/contract.rs
keyleu:
```rust
pub struct HandleMsg {
    pub origin: u32,
    pub sender: HexBinary,
    pub body: HexBinary,
}
```

Someone can send anything it wants in the body of the handleMsg. 
Here we are checking that the mailbox is correct, that it comes from the domain_id we registered (not another EVM), and that the address that sent it is the address of the processor we registered. These 3 things should benough to validate it

contracts/authorization/src/contract.rs
keyleu: kinda hate updates on maps tbh, looks very ugly and hard to read

packages/bridging-utils/Cargo.toml
keyleu: renamed

packages/authorization-utils/src/domain.rs
keyleu: changed

contracts/authorization/src/authorization.rs
```diff
@@ -80,37 +92,164 @@ impl Validate for Authorization {
...
+            Domain::External(name) => Some(
+                EXTERNAL_DOMAINS
+                    .load(store, name.clone())
+                    .map_err(|_| ContractError::DomainIsNotRegistered(name.clone()))?,
+            ),
+        };
+
+        for func in functions {
+            self.validate_domain(func, domain)?;
+            self.validate_message_type(func, &external_domain)?;
```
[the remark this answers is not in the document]
keyleu: changed the structure of validate_message_type a little to have it make more sense

contracts/authorization/src/authorization.rs
```diff
@@ -64,8 +72,12 @@ impl Validate for Authorization {
...
 
         match &self.subroutine {
-            Subroutine::Atomic(config) => self.validate_functions(store, &config.functions)?,
-            Subroutine::NonAtomic(config) => self.validate_functions(store, &config.functions)?,
+            Subroutine::Atomic(config) => {
+                self.validate_functions(store, &config.functions, querier)?
+            }
+            Subroutine::NonAtomic(config) => {
+                self.validate_functions(store, &config.functions, querier)?
+            }
```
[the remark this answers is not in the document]
keyleu: its kinda not that simple, it's more an authorization validation because it also validates things like if the domain already exists in the contract. Also the calls are not identical because the functions in each subroutine are also different. What I did tho is add a wrapper around this function to make it more clear.

keyleu:
Applied most changes asked @bekauz and @Art3miX, which were mostly nits, going to proceed to start writing tests from this

packages/authorization-utils/src/msg.rs
Art3miX:
```impl ExternalDomainInfo {
    pub fn to_external_domain_validated(&self, api: &dyn Api) -> StdResult<ExternalDomain> {
        Ok(ExternalDomain {
            name: self.name.clone(),
            execution_environment: self.execution_environment.clone(),
            connector: crate::domain::Connector::PolytoneNote {
                address: self.connector.to_addr(api)?,
                timeout_seconds: self.connector.timeout_seconds(),
                state: crate::domain::PolytoneProxyState::PendingResponse,
            },
            processor: self.processor.clone(),
            callback_proxy: crate::domain::CallbackProxy::PolytoneProxy(
                self.callback_proxy.to_addr(api)?,
            ),
        })
    pub fn to_external_domain_validated(self, api: &dyn Api) -> StdResult<ExternalDomain> {
        self.into_external_domain(api)
    }
}

You also have this block of code.

File: contracts/authorization/src/contract.rs (hunk 8 above)
Comment (Art3miX) [reply contracts/authorization/src/contract.rs:1158]: Oh ok, its worth mentioning that in the comment of this function to make sure its clear that handle_msg is sent by the mailbox, and the body is the callback data.

File: packages/authorization-utils/src/msg.rs (hunk 6 above)
Comment (keyleu) [reply packages/authorization-utils/src/msg.rs]: Right I see, combined it together, didn't realize I accidentally separated.

File: contracts/authorization/src/contract.rs (hunk 8 above)
Comment (keyleu) [reply contracts/authorization/src/contract.rs:1159]: added a comment clarifying it

**Comment by keyleu** [comment]
Added tests for all the functionality. Don't want to bloat this PR more so if you think there is any test missing/worth adding, I'll create an issue and add it in a separate PR. I'll also add the EVM integration tests to CI in a separate PR as it might not work out of the box.

File: local-interchaintest/examples/ethereum_integration_tests.rs
Comment (bekauz) [inline local-interchaintest/examples/ethereum_integration_tests.rs:1]: overall this makes perfect sense to me. love the idea of blocking the main thread until some call completes instead of sleeping; perhaps now with more examples of how integration tests are done on different domains we can find a way to standardize this somehow..

out of curiosity, why tokio `block_on` over native `futures::executor::block_on`?

File: packages/authorization-utils/src/domain.rs [hunk 15]
```diff
@@ -11,54 +11,88 @@ pub enum Domain {
...
-        match &self.connector {
-            Connector::PolytoneNote { state, .. } => state.clone(),
+    pub fn get_polytone_proxy_state(&self) -> Option<PolytoneProxyState> {
+        match self {
+            ExecutionEnvironment::Cosmwasm(CosmwasmBridge::Polytone(polytone_info)) => {
+                Some(polytone_info.polytone_note.state.clone())
+            }
+            ExecutionEnvironment::Evm(_, _) => None,
         }
     }
```
bekauz, line 53: I feel like having this under `impl ExecutionEnvironment` makes it a little confusing, what about we move it to `CosmwasmBridge` or `PolytoneConnectors`?  seems like it's only called under `PolytoneCallbackMsg::CreateProxy` handler so we already know we're dealing with polytone there so shouldn't be much issues

packages/authorization-utils/src/domain.rs
```diff
@@ -11,54 +11,88 @@ pub enum Domain {
...
+        match self {
+            ExecutionEnvironment::Cosmwasm(CosmwasmBridge::Polytone(polytone_info)) => {
+                polytone_info.polytone_note.state = state;
+                Ok(())
             }
+            ExecutionEnvironment::Evm(_, _) => Err(StdError::generic_err(
+                "EVM domain does not have a polytone proxy state",
+            )),
+        }
+    }
```
bekauz, line 65: same here, I think it would be cleaner if we moved this impl away from `ExecutionEnvironment` and into something more specific

packages/authorization-utils/src/domain.rs
```diff
@@ -11,54 +11,88 @@ pub enum Domain {
...
+
+    pub fn get_encoder(&self) -> Option<&Encoder> {
```
bekauz, line 67: is this used anywhere?

packages/authorization-utils/src/msg.rs
```diff
@@ -93,6 +163,9 @@ pub enum ExecuteMsg {
     // Polytone callback listener
     #[serde(rename = "callback")]
     PolytoneCallback(CallbackMessage),
+    // Hyperlane callback listener
+    #[serde(rename = "handle")]
```
bekauz, line 167: now I get why in the authorizations contract execute match statement on the msg you had `HyperlaneCallback(*handle_msg*)` while `PolytoneCallback(*callback_msg*)`. feel free to disregard that one!

packages/authorization-utils/src/msg.rs
```diff
@@ -193,42 +266,60 @@ pub enum InternalAuthorizationMsg {
...
+                ProcessorMessage::CosmwasmExecuteMsg { .. },
+                MessageType::CosmwasmExecuteMsg
+            ) | (
+                ProcessorMessage::CosmwasmMigrateMsg { .. },
+                MessageType::CosmwasmMigrateMsg
+            ) | (ProcessorMessage::EvmCall { .. }, MessageType::EvmCall(..))
+                | (ProcessorMessage::EvmRawCall { .. }, MessageType::EvmRawCall)
+        )
     }
+}
```
bekauz, line 287: maybe worth unit testing impls like this?

bekauz (commented):
did another pass and overall it looks great! left a few points and questions.

on a high level I love that more logic is moving into utils which are free of `Deps/DepsMut`. things are getting easier to be unit tested as there is no need to rebuild wasms after every change + codecov.

contracts/authorization/src/contract.rs
```diff
@@ -158,6 +169,9 @@ pub fn execute(
         ExecuteMsg::PolytoneCallback(callback_msg) => {
             process_polytone_callback(deps, env, info, callback_msg)
         }
+        ExecuteMsg::HyperlaneCallback(handle_msg) => {
+            process_hyperlane_callback(deps, env, info, handle_msg)
+        }
```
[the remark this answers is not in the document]
keyleu: kinda prefer to keep this one as handle_msg as we actually receive a HandleMsg

contracts/authorization/src/contract.rs
```diff
@@ -953,6 +1120,97 @@ fn process_polytone_callback(
...
+                        Err(_) => None,
+                    }
+                }
+                // Return None for non-EVM/Hyperlane execution environments
+                _ => None,
+            }
+        })
+        .ok_or(ContractError::Unauthorized(
+            UnauthorizedReason::UnauthorizedCallbackSender {},
+        ))?;
```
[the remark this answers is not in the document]
keyleu: I don't think it's worth it, this will never error unless someone is actually trying to attack the contract so in the end it will be an unauthorized sender. If a message is received and for some reason there is not a single domain that matches all those conditions in our state, it means some unauthorized party is sending us something that we shouldn't process, it kinda needs to match all the conditions.

local-interchaintest/examples/ethereum_integration_tests.rs
keyleu: no reason tbh, started using tokio block_on for other things and just kept using it to not change

packages/authorization-utils/src/domain.rs
keyleu: Agree, moved it to PolytoneConnectors

packages/authorization-utils/src/domain.rs
keyleu: no, removed it.

packages/authorization-utils/src/msg.rs
keyleu: Added!

contracts/authorization/src/contract.rs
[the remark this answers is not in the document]
bekauz: I see what you mean. I thought that this may be legitimately misconfigured in some way, e.g. if the program runs across multiple domains maybe some mix up in the hyperlane config could lead to some (but not all) of those checks failing.

bekauz (approved):

PR #160, opened 2025-01-22
[processor] add execution time limit for subroutines

### LINKED ISSUE #137: Add an execution time limit to subroutines
Add a timestamp limit for subroutines to be executed in the processor. If a subroutine hasn't been executed after X time, it will be expired and stop executing

najeal:
Closes #137 

This is a proposition for adding execution time limit of the subroutines. It adds a const in config for `timeout` but it can be provided another way.
Is there a specific way to test locally ?

### DIFF
--- a/contracts/authorization/src/tests/processor.rs
+++ b/contracts/authorization/src/tests/processor.rs
@@ -2023,6 +2023,9 @@ fn failed_atomic_batch_after_retries() {
                     }),
                     priority: Priority::Medium,
                     retry: None,
+                    expiration_time: cw_utils::Expiration::AtTime(
+                        setup.app.get_block_timestamp().plus_seconds(100),
+                    ),
                 },
             }),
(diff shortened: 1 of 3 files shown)


keyleu:
Hey @najeal, thanks for this. We are currently in the middle of EVM integration so this change actually involves more things for both environments and would come after that.
For example there needs to be a different ExecutionResult for these `expired` batches that needs to be relayed to the authorization contract, also removing it from the queue not just returning or the queue would be stuck with that expired batch. Let's wait trying to fix this until EVM integration is done.

In any case, for testing it locally we are using test-tube for single domain and local-interchaintest for multidomain, you have examples on how to run the latter in the local-interchaintest folder

najeal:
@keyleu thanks for your feedback, I will close this PR so 👍 
If you see any other issue I could take a look, let me know !

PR #162, opened 2025-01-23
Intro edit

hxrts:
small adjustment, mainly because I'm sharing the intro link publicly now


hxrts:
oh weird, not sure why this branch is dirty, with that light mode change. should I make a new PR?

Issue #152, opened 2025-01-16
[docs] accounts section


bekauz:
currently our docs only describe accounts under the `/components` section, providing a high level overview of what is the role of accounts in Valence Protocol.

Given that now we also have the notion of *storage accounts*, it would be helpful to create a dedicated docs section meant entirely for accounts, sitting at the same level and following the structure of `/libraries` and `authorizations_processors`.

PR #181, opened 2025-01-31
asserter docs

bekauz:
# description

base implementation of Valence Asserter component documentation


docs/src/components/_overview.md
```diff
@@ -6,6 +6,7 @@ There are two ways to execute Valence Programs.
 Valence currently supports CosmWasm and EVM. SVM support coming soon. The rest of this section provides a high-level breakdown of the components that comprise a Valence Program using on-chain coprocessors.
     - [Domains](./domains.md)
     - [Accounts](./accounts.md)
+    - [Storage Accounts](./storage_account.md)
```
keyleu, line 9: imo we should have subsections under accounts and there have base account, storage account, ICA account etc.
Either that or rename accounts above to base_accounts because account is too generic

docs/src/components/storage_account.md
```diff
@@ -21,32 +21,34 @@ pub enum ExecuteMsg {
...
 
-If `PostData` is called by the owner or an approved library, it will persist the *key-value*
+If `StoreValenceType` is called by the owner or an approved library, it will persist the *key-value*
 mapping in its state. Storage here works in an overriding manner, meaning that posting data
 for a key that already exists will override its previous value and act as an update method.
 
 ### Query Methods
 
-Once data had been posted into the storage account using `PostData` call, it is made available
+Once a given type had been posted into the storage account using `StoreValenceType` call, it becomes available
```
keyleu, line 40:
```suggestion
Once a given type has been posted into the storage account using `StoreValenceType` call, it becomes available

docs/src/components/_overview.md
bekauz: good point. for now I clarified all previous references to `Accounts` to be `Base Accounts` as there is already an issue for a dedicated accounts section: #152 . I updated it to also include the *ICA Accounts* section there.

keyleu (approved):

uditvira (approved):

PR #188, opened 2025-02-01
Fix broken mermaid diagram

uditvira:


hxrts (commented):
I see, object names need to be unbroken strings

hxrts (approved):

PR #191, opened 2025-02-03
manager server changes

Art3miX:
This includes changes needed for the server to work, will need to update it with current changes as its fairly old.

## Edit:
Updated from dev
Made sure its working


local-interchaintest/examples/manager_server_dev.rs
bekauz, line 1: couldn't this be done inside `setup_manager()` call itself?

Art3miX: I assume you mean the test itself.

I made this test to "export" the global config the manager uses to a file, the reason we want this is because this is the exact config we need for working with the manager server (both prod and dev).
This is super useful for using and testing the manager server (different repo) with the interchaintest chains, you can start your chains, run this test, and you get a ready to use global config file which you simply copy into the server and you ready to go.

program-manager/src/domain/cosmos_cw.rs
```diff
@@ -771,21 +771,24 @@ impl Connector for CosmosCosmwasmConnector {
...
+        for library in config.libraries.values_mut() {
             if library.addr.is_none() {
                 return Err(CosmosCosmwasmError::Error(anyhow::anyhow!(
                     "Before saving program config each library must have an address"
                 ))
                 .into());
             }
+
+            // Set config to None before saving on-chain
+            library.config = LibraryConfig::None;
```
keyleu, line 791: just 1 question? why are we doing this?

Art3miX: We decided to not save the configs in the registry, it used to be "parsed off" when parsing to JSON, but it caused an issue so I basically set it to None when we save the config to the registry.

keyleu (approved):

Issue #199, opened 2025-02-07
middleware asserter


bekauz:
# description

implement a predicate asserter contract that would be capable of evaluating basic assertions regarding two values.

values should be configurable as either constants of variables. variable values should be fetched from a storage account and accessed over `ValenceType` query implementations.

evaluation function should:
- return an error in case the assertion evaluates to false
- return a successful response in case the assertion evaluates to true

Issue #192, opened 2025-02-04
fix instantiate2 address prediction in local-interchain


bekauz:
# description

proxy address prediction done in local-interchaintest authorization utils is currently broken for osmosis (and possibly other domains).

this is due to how `localic_utils::utils::queries::TestContextQuery` and local-ic std formats certain requests. in address prediction case, flags that are only recognized by certain chains are appended to the request, leading to `connect: connection refused` errors.

bekauz:
closed with #200

PR #200, opened 2025-02-07
middleware asserter

### LINKED ISSUE #153: neutron IC querier follow-up

bekauz:
closes #153 #199 #192 

# description

introduces Valence Asserter which enables conditional execution of processor batches.

## e2e

in addition to unit testing, asserter functionality in Valence Programs is verified with an e2e test for the following program:

> provide liquidity on a remote domain only if the liquidity depth of a particular denom in a given dex pool is below a certain threshold

this is achieved as follows:

0. setup the env (osmosis + neutron, icq relayer, middleware)
1. register KV query targeting the specified pool on osmosis
2. set up conditional IBC forwarding of the target denom with which liquidity is to be provided. this IBC forwarding happens if and only if the asserter confirms that the amount of this denom in the pool being queried via step 1. is below a certain threshold.
3. provide single-sided liquidity on osmosis

## other changes

in addition to above, this pr:
- adds `GetPoolAssetAmount { target_denom: String }` query on `ValenceXykPool`
- reworks the `set_up_external_domain_with_polytone` function on local-ic authorization utils (#192)
- updates `neutron-ic-querier` to check for query registration deposit with a self bank query
- extends local-interchaintest readme with neutron icq relayer setup instructions


contracts/middleware/asserter/src/contract.rs
```diff
@@ -0,0 +1,92 @@
...
+    // first we fetch the values we want to compare
+    let a_cmp = get_comparable_value(deps.querier, a)?;
+    let b_cmp = get_comparable_value(deps.querier, b)?;
+
+    match (a_cmp, b_cmp) {
+        (ValencePrimitive::Decimal(a), ValencePrimitive::Decimal(b)) => Ok(predicate.eval(a, b)),
+        (ValencePrimitive::Uint256(a), ValencePrimitive::Uint256(b)) => Ok(predicate.eval(a, b)),
+        (ValencePrimitive::Uint128(a), ValencePrimitive::Uint128(b)) => Ok(predicate.eval(a, b)),
+        (ValencePrimitive::Uint64(a), ValencePrimitive::Uint64(b)) => Ok(predicate.eval(a, b)),
+        (ValencePrimitive::String(a), ValencePrimitive::String(b)) => Ok(predicate.eval(a, b)),
```
keyleu, line 65: Are we getting the desired behavior for Strings here? Strings comparisons in rust do a byte per byte value. E.g. 'b' would be greater than 'Shoe' for example. I guess for Strings 99% of use cases will be checking if they are equal

local-interchaintest/examples/conditional_lp.rs
```diff
@@ -0,0 +1,1036 @@
...
+    let storage_accounts = create_storage_accounts(
+        test_ctx,
+        DEFAULT_KEY,
+        NEUTRON_CHAIN_NAME,
+        storage_acc_code_id,
+        NEUTRON_CHAIN_ADMIN_ADDR.to_string(),
+        vec![
+            neutron_processor_address.to_string(),
+            NEUTRON_CHAIN_ADMIN_ADDR.to_string(),
+        ],
```
keyleu, line 542: just to confirm, this is not being used right? Removing these 2 addresses from approved libraries wouldn't fail the test I mean

contracts/middleware/asserter/Cargo.toml
```diff
@@ -0,0 +1,37 @@
...
+prost-types              = { workspace = true }
+neutron-sdk              = { workspace = true }
+serde-json-wasm          = { version = "1.0.0" }
+thiserror                = { workspace = true }
+serde_json               = { workspace = true }
+cw-storage-plus          = { workspace = true }
+valence-middleware-utils = { workspace = true }
+semver                   = { workspace = true }
+cw2                      = { workspace = true }
+valence-storage-account  = { workspace = true }
```
keyleu, line 30: was a bit confused when looking at this. Most of these are not being used right, I guess it got copy pasted. Can we remove all that is not being used here?

local-interchaintest/src/utils/authorization.rs
```diff
@@ -401,32 +412,30 @@ pub fn set_up_external_domain_with_polytone(
...
+    let processor_instantiation_str = format!(
+        "tx wasm instantiate2 {external_processor_code_id} {} {} \
+        --label valence_processor \
+        --chain-id={chain_id} \
+        --no-admin \
+        {extra_flags} \
+        --from {DEFAULT_KEY}",
+        serde_json::to_value(processor_instantiate_msg)?,
+        salt,
+    );
```
keyleu, line 431: was build_tx_instantaite2 not working?

local-interchaintest/examples/conditional_lp.rs
```diff
@@ -0,0 +1,1036 @@
...
+            AtomicSubroutineBuilder::new()
+                .with_function(
+                    AtomicFunctionBuilder::new()
+                        .with_domain(valence_authorization_utils::domain::Domain::Main)
+                        .with_contract_address(LibraryAccountType::Addr(asserter.clone()))
+                        .with_message_details(MessageDetails {
+                            message_type: MessageType::CosmwasmExecuteMsg,
+                            message: Message {
+                                name: "assert".to_string(),
+                                params_restrictions: None,
```
keyleu, line 787: totally optional, but to make this test 100% real you can restrict the json sent in the assertion here to match the  exact assertion, this way the user cannot just send anything he wants here.

keyleu:
Very complete local-ic tests btw! Very nice seeing all the pieces working together as expected

contracts/middleware/asserter/Cargo.toml
```diff
@@ -0,0 +1,37 @@
...
+
+[dev-dependencies]
+cw-multi-test              = { workspace = true }
+valence-middleware-osmosis = { workspace = true }
+osmosis-std                = { workspace = true }
+osmosis-std-derive         = { workspace = true }
+valence-account-utils     = { workspace = true }
```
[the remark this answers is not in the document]
bekauz: fixed.

thoughts on adding a preconfigured formatter? I played around with one on [this branch](https://github.com/timewave-computer/valence-protocol/tree/benskey/toml-formatter) and it seems to work. Wasn't sure if it was necessary.

contracts/middleware/asserter/README.md
```diff
@@ -0,0 +1,114 @@
...
+```
+
+where
+
+```rust
+pub struct QueryInfo {
+    // addr of the storage account
+    pub storage_account: String,
+    // key to access the value in the storage account
+    pub storage_slot_key: String,
```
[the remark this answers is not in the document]
bekauz: suppose we could name it simply `storage_key`. idea behind specifying *slot* was to show that there can be multiple keys stored under the same storage account.

contracts/middleware/asserter/src/contract.rs
bekauz: I imagine that in practice string comparisons will only deal with `EQ` predicates. To keep the api consistent I kept the logic identical to that of comparing any other types.

I have the following unit tests for it, do you think anything is missing?

```rust

#[test]
fn double_const_string_assertion_lt_happy() {
    let mut suite = Suite::default();

    assert!(suite
        .assert(
            AssertionValue::Constant(ValencePrimitive::String("abc".to_string())),
            Predicate::LT,
            AssertionValue::Constant(ValencePrimitive::String("def".to_string())),
        )
        .is_ok())
}

#[test]
fn double_const_string_assertion_eq_happy() {
    let mut suite = Suite::default();

    assert!(suite
        .assert(
            AssertionValue::Constant(ValencePrimitive::String("abc".to_string())),
            Predicate::EQ,
            AssertionValue::Constant(ValencePrimitive::String("abc".to_string())),
        )
        .is_ok())
}

#[test]
#[should_panic(expected = "assertion failed")]
fn double_const_string_assertion_err() {
    let mut suite = Suite::default();

    assert!(suite
        .assert(
            AssertionValue::Constant(ValencePrimitive::String("a".to_string())),
            Predicate::GT,
            AssertionValue::Constant(ValencePrimitive::String("def".to_string())),
        )
        .is_ok())
}
```

local-interchaintest/src/utils/authorization.rs
bekauz: unfortunately for osmosis (and probably more domains that we haven't confirmed yet), it was not: #192

contracts/middleware/asserter/README.md
```diff
@@ -0,0 +1,114 @@
...
+    GT,
+    GTE,
+}
+```
+
+Assertion values can be specified with the following type:
+
+```rust
+pub enum AssertionValue {
+    // storage account slot query
```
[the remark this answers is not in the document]
keyleu: nah was just worried we might confuse it with slots in EVM state, I don't think it's a big deal

contracts/middleware/asserter/src/contract.rs
keyleu: unit tests makes sense, just saying anything that is not EQ in String has a behavior that might be unexpected to some people, like maybe they think it's alphabetical or length based. As long as they are familiar with string comparisons they should be fine

local-interchaintest/src/utils/authorization.rs
keyleu: that was the predict query though right?, I'm surprised instantiate2tx builder was not working

contracts/libraries/neutron-ic-querier/src/contract.rs
```diff
@@ -139,11 +139,21 @@ fn register_kv_query(
...
-
+        for coin in info.funds.iter() {
+            // if the sender attached this fee denom, we validate that the amount is sufficient
+            if coin.denom == fee_coin.denom {
+                ensure!(
+                    coin.amount >= fee_coin.amount,
+                    StdError::generic_err("insufficient icq registration fee amount")
+                );
+            }
+        }
```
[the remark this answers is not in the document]
keyleu: Right, if it's an array of funds then yeah this def makes more sense, thought the fee would be only 1 coin. Nvm ignore the suggestion then!

local-interchaintest/src/utils/authorization.rs
bekauz: you're right, this was a leftover from debugging. reverted to using `build_tx_instantaite2`.

local-interchaintest/examples/conditional_lp.rs
bekauz: correct, emptied the array to avoid confusion

contracts/middleware/asserter/src/contract.rs
bekauz: yeah, totally agree that lexicographic ordering might behave not as expected to some users, going to highlight that in the docs. That being said, I can imagine this being useful for some checks that go beyond basic EQ, e.g. comparing ISO timestamps stored as strings.

For now I added the following comment:

```rust
        // note: string predicates are evaluated in lexicographical order and are case-sensitive:
        // - "Z" < "a"
        // - "Assertion" != "assertion"
        (ValencePrimitive::String(a), ValencePrimitive::String(b)) => Ok(predicate.eval(a, b)),
```

keyleu (approved):
I think only one of the issues will close automatically, remember to close the other two after merging

PR #204, opened 2025-02-10
Adding doc and troubleshooting section for running program tests

elenamik:
Adding a doc to prevent the error I was encountering running the example tests, and proposing a troubleshooting section. I am not sure if it is the best place for one, but could be nice to have a resource listing possible reasons for common errors.


local-interchaintest/README.md
```diff
@@ -18,18 +18,42 @@ local-ic start neutron_juno --api-port 42069
...
 
+Check that the `local-ic start` command created `local-interchaintest/configs/logs.json` with the correct RPC URL mappings. This configuration file is required to run the example tests.
+
 Once you have your tests written, you can run them using the following command from the workspace directory, here I'm running the `polytone` tests that are in the `examples` folder:
 
 ```bash
 cargo run --package local-interchaintest --example polytone
 ```
+
+## Troubleshooting
```
elenamik, line 45: not sure where is the best place for something like this to live, or if it is something we can maintain, but could be a helpful resource to new users of the tools?

local-interchaintest/README.md
```diff
@@ -18,18 +18,42 @@ local-ic start neutron_juno --api-port 42069
...
+Error: LocalInterchain(Custom { msg: "channel_json is not an array" })
+```
+The chains required in the test are not running.
+
+`cargo run --package local-interchaintest --example example_file_name`
+```txt
+called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }
+note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
+```
+Check that `local-interchaintest/configs/logs.json` exists. If it does not, kill all local-ic processes and rerun from `local-interchaintest` directory.
```
Pacman99, line 59: This error could also happen if contracts weren't built

Pacman99 (approved):

local-interchaintest/README.md
elenamik: kk, if we are down to maintain a troubleshooting section here, I can add this as well. I don't know if it makes more sense to add this in to centralized docs resource instead?

local-interchaintest/README.md
```diff
@@ -18,18 +18,40 @@ local-ic start neutron_juno --api-port 42069
...
```
keyleu, line 57: re-running local-ic start ...  restarts everything so no need to kill manually I think

elenamik: its not 100% guaranteed that everything is terminated in docker

local-interchaintest/README.md
keyleu: I think adding it here is fine

local-interchaintest/README.md
keyleu: maybe add the command to kill local-ic process and rephrase a little: mention what needs to be rerun

local-interchaintest/README.md
Art3miX: The problem is its a generic error for any action that trying to read a file but it doesn't exists.
The better solution would be to make the error more specific regarding what doesn't exists.

Art3miX (approved):

keyleu (approved):

PR #208, opened 2025-02-11
Bump workspace version

keyleu:
Bump version to 0.2.0


keyleu:
Some contracts were not respecting versioning correctly, so checked all of them and updated it to take it from workspace (except singletons like the encoder). Also removed all hardcoding of authors which was inconsistent.

bekauz (approved):

PR #212, opened 2025-02-12
docs: change deployment folder

keyleu:
When adding linkcheck we have multiple outputs so they are created in subdirectories.
Therefore we must deploy the one in html


hxrts (approved):
LGTM

PR #214, opened 2025-02-12
deployment scripts

Art3miX:
Added scripts to make it easier to deploy programs using the manager
Added docs for them as well.


docs/src/deployment/local.md
```diff
@@ -0,0 +1,77 @@
...
+
+Use CosmWasm optimizer to optimize contracts and store the results in `./artifacts`
+
+```bash
+just optimize
+```
+
+Or
+
+```base
```
keyleu, line 33:
```suggestion
```bash

File: docs/src/deployment/local.md [hunk 2]
```diff
@@ -0,0 +1,77 @@
...
+
+Example with `neutron_juno.json` chain config:
+
+```bash
+cargo run -p generate_local_ic_config -- -c neutron_juno
+```
+
+## 5. Deploy a program
+
+To deploy a program, you can use the `deploy_program` script. In (deploy_program)[deployment/scripts/deploy_program/src/main.rs] you will find an example to a built program using our program builder, you can customize this to build your own program.
```
keyleu, line 55:
```suggestion
To deploy a program, you can use the `deploy_program` script. In (deploy_program)[deployment/scripts/deploy_program/src/main.rs] you will find an example of a built program using our program builder, you can customize this to build your own program.

File: docs/src/deployment/local.md [hunk 3]
```diff
@@ -0,0 +1,77 @@
...
+
+After you customized your program, you can deploy it using the following command:
```
keyleu, line 57:
```suggestion
After you customize your program, you can deploy it using the following command:

File: docs/src/deployment/local.md [hunk 4]
```diff
@@ -0,0 +1,77 @@
...
+
+```bash
+cargo run -p deploy_program
+```
+
+By default this will deploy a program to the local enviroment, you can change this by passing the `-c` argument with the wanted enviroment, for example:
```
keyleu, line 63:
```suggestion
By default this will deploy a program to the local environment, you can change this by passing the `-c` argument with the wanted enviroment, for example:

File: docs/src/deployment/local.md [hunk 5]
```diff
@@ -0,0 +1,77 @@
...
+
+```bash
+cargo run -p deploy_program -- -c mainnet
+```
+
+Options are: `local`, `mainnet` and `testnet`.
```
keyleu, line 69: Right now we are only supporting local right? Maybe remove other options for now?

keyleu (dismissed):

deployment/configs/mainnet/chains.toml
```diff
@@ -0,0 +1,30 @@
...
+
+[chains.neutron]
+name      = "neutron"
+rpc       = "https://neutron-testnet-rpc.polkachu.com:443"
+grpc      = "https://grpc-falcron.pion-1.ntrn.tech:80"
+# grpc      = "https://neutron-testnet-rpc.polkachu.com:19190"
+prefix    = "neutron"
+gas_price = "0.055"
+gas_denom = "untrn"
+coin_type = "118"
```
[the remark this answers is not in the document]
Art3miX: mainnet and testnet configs are just placeholders for now to have them, we didn't deploy to either of those yet.
This is also the format the manager is expecting so its easier to build the CI to update those configs as we go.

deployment/configs/mainnet/contracts.toml
```diff
@@ -0,0 +1,17 @@
...
+valence_splitter         = 6049
+valence_reverse_splitter = 3
+
+[contracts.code_ids.cosmos-hub]
+valence_authorization    = 7217
+valence_processor        = 7218
+valence_base_account     = 7219
+valence_forwarder        = 7222
+valence_splitter         = 6049
+valence_reverse_splitter = 3
```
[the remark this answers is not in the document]
Art3miX: Same as above

docs/src/deployment/_overview.md
```diff
@@ -0,0 +1,6 @@
+# Deployment
+
+This section contains a detailed explanation of how to deploy programs on different enviroments.
```
keyleu, line 3:
```suggestion
This section contains a detailed explanation of how to deploy programs on different environments.

**Comment by keyleu** [comment]
can we also merge main to take updated docs CI?

**Review by bekauz — COMMENTED** [review]
some high level thoughts:

first, it would be very helpful to have more documentation with regards to the overall lifecycle of these deployment scripts and how they fit in to the rest of the flow. was thinking mermaid diagrams or any other means to highlight the order of operations, dependencies between different steps, etc. 

somewhat related to that, I think we should be more explicit with describing the steps needed for users to customize and test their own programs. For instance, looking at the `generate_local_ic_config` it is quite hard to tell which of the contracts being deployed are critical for overall protocol and which are program-specific.
following up on the config generation, I think it may be a good idea to strip away all boilerplate from these files. if all ic config generation scripts are expected to export the manager config in the end, maybe we can abstract that away from the users and call it elsewhere to reduce the number of things developers need to worry about? just brainstorming here.

lastly, maybe it's worth to first get the localhost deployment working well before setting up the testnet/mainnet ones? trying to get all of them working in one go is quite a lot, and I think by polishing up the localhost logic first we would simplify the remaining lift needed for enabling mainnet and testnet functionality.

File: deployment/configs/mainnet/neutron_bridges/neutron_cosmos_hub.toml [hunk 9]
```diff
@@ -0,0 +1,12 @@
+[bridges.neutron.cosmos_hub.polytone.neutron]
+voice_addr      = "neutron1vq2z7zgj7v4yj7z2zgj7v4yj7z2zgj7z2zgj7v"
+note_addr       = "neutron1vq2z7zgj7v4yj7z2zgj7v4yj7z2zgj7z2zgj7v"
+other_note_port = "wasm.cosmos1vq2z7zgj7v4yj7z2zgj7v4yj7z2zgj7z2zgj7v"
+connection_id   = "connection-0"
+channel_id      = "channel-0"
+[bridges.neutron.cosmos_hub.polytone.cosmos_hub]
```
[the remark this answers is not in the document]
Art3miX: How would that look in a rust type?

deployment/scripts/deploy_program/src/main.rs
```diff
@@ -0,0 +1,227 @@
...
+                            None
+                        } else {
+                            Some(config::File::from(p))
+                        }
+                    })
+                    .collect::<Vec<_>>(),
+            )
+            .build()?
+            .try_deserialize()
+            .map_err(|e| e.into()),
```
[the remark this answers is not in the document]
Art3miX: Reading the config is pretty specific to the environment, they currently share the same structure but doesn't have to be the case, and might not be the case in the future.

deployment/scripts/generate_local_ic_config/src/main.rs
```diff
@@ -0,0 +1,74 @@
...
+            ASTROPORT_LPER_NAME,
+            ASTROPORT_WITHDRAWER_NAME,
+            OSMOSIS_GAMM_LPER_NAME,
+            OSMOSIS_GAMM_LWER_NAME,
+            OSMOSIS_CL_LPER_NAME,
+            OSMOSIS_CL_LWER_NAME,
+            POLYTONE_NOTE_NAME,
+            POLYTONE_VOICE_NAME,
+            POLYTONE_PROXY_NAME,
+        ],
```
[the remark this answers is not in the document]
Art3miX: I'm not sure there is a single best way here, this is the reason we specify which contracts to upload in the first place.
In our own tests, we want to be more specific what contracts to upload where to reduce unneeded uploads, so we do this per test which is fine.

But thinking of the user, we basically want everything to be available because we have no idea what their specific program is.

We can read the program, and generate the manager config based on the domains included in the program, but my idea is that local-ic is running independent of the program, doing this will tie the local-ic chains to a single program which I think not what we want.

I think uploading everything to all chains will work the best as a starting point, we can expend this later and allow users to optimize it a little more for their use case.

Regarding contracts which only works on a single chain like the osmosis example, this is a good point, but again, this is somewhat hard to implement, what if in the future we would have a contract for a chain feature? where a contract will only work if a feature exists on a chain, how would you know if a chain supports this feature or not?
This is something we will also have to think if we will have automated deployment from CI, obviously we don't need to upload osmosis specific thing on Neutron, but how would we know? 

The simplest why would be by name, if a contract contains a chain name we only upload it to that chain, but this only works for some cases, im sure we don't want contract names like: `neutron_osmosis_juno_splitter`, we also need a list of chain names for this to know what chain names to look for in names any way.
Other way is to have a mapper between a chain and a contract, but this might get hard to maintain with more libraries.

For conclusion, I think it is best to upload all contracts to all chains for local development because this step is only needed once per chain spin up, and we have no info regarding the users use case to determine what contracts to upload.
Relying on contracts names is fine for now, but can get really complicated.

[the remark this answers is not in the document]
Art3miX: Thinking about it now, best way would be to create a chain image with all contracts uploaded already, we will have to maintain it with new chain versions which can get complicated, and a whole system is needed to make this possible.
But this will enable everything right away when you spin up the chains without this extra upload step.

This is also not a big deal if we move local-ic into a persistent server, this is only needed once the environment is restarted.

docs/src/deployment/local.md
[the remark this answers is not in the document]
Art3miX: Not sure what audience this should be for if not for developers. 
For non-developers we want to have a UI and docs on how to use the UI.
To use this PR scripts the users need to write rust code (program builder), plus they need to run local-ic locally for local development, this is not the experience non-developers expect for.

deployment/configs/mainnet/contracts.toml
```diff
@@ -0,0 +1,17 @@
...
+
+[contracts.code_ids.neutron]
+valence_authorization    = 7217
+valence_processor        = 7218
+valence_base_account     = 7219
+valence_forwarder        = 7222
+valence_splitter         = 6049
+valence_reverse_splitter = 3
+
+[contracts.code_ids.cosmos-hub]
```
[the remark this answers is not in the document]
Art3miX: This is mainnet config so not active yet.
As long as we choose a single chain name doesn't matter, this basically corresponds to the domain name we will give to it.

docs/src/deployment/local.md
```diff
@@ -0,0 +1,91 @@
...
+In order to test a program locally, we use the local interchaintest suite to spin up chains.
+
+## 1. Installing local-interchain
+
+Before you can run the tests, you need to install local interchain. This is a one-time operation. NOTE: your binary will link back to the location where you install, if you remove the folder, you need to run `make install` again.
+
+```bash
+git clone https://github.com/strangelove-ventures/interchaintest && cd interchaintest/local-interchain && make install
+```
+
```
keyleu, line 12: you also need to build the solidity artifacts as well, so there is a step missing here before running local-ic that involves:

1) Installing Foundry
2) Installing dependencies (forge soldeer install in /solidity)
3) Building contracts (forge build in solidity/)

docs/src/deployment/local.md
```diff
@@ -0,0 +1,91 @@
...
+## 2. Running your local environment
+
+Run one of the set-up configs we have in the `local-interchaintest/chains` folder. For example, to run the `neutron.json` config, run the following command inside the `local-interchaintest` folder:
```
keyleu, line 15:
```suggestion
Run one of the set-up configs we have in the `e2e/chains` folder. For example, to run the `neutron.json` config, run the following command inside the `e2e` folder:

docs/src/deployment/local.md
Art3miX: Currently Ethereum is not supported for the users, I suggest to separate this some how so users that need Ethereum will need to do extra steps instead of forcing everyone to do this.

keyleu: We can't separate it, you need to have the contracts at compile time in Rust to have the types, is not something we can do at runtime.

keyleu: discussed it

keyleu (approved):

Issue #205, opened 2025-02-11
Timeouts two ways


hxrts:
There are two remote timeout constructions we discussed. The first of which needs to be avoided, the second we need to be careful with.

**1) Program starts on Chain A, advances to Chain B, and you want to reason on Chain A whether remote execution timed out in order to initiate a continuation.**
What you DO NOT want to do here is use the Chain A clock time as a proxy for the Chain B clock and presume that the absence of a timeout delivery packet implies a remote time out occurred. This could lead to a situation where the packet is delivered and accepted on Chain B, but due to clock drift or network latency the logic on Chain A draws the wrong conclusion. Ex post inference about remote *evaluation results* must be avoided.

**2) Program starts on Chain A, the relayer goes down for some indefinite amount of time, after which it restarts message delivery. When the packet is delivered on Chain B you want to reason about timeout expiry.**
This is the scenario relevant to Keyne's PR, and the construction we're using is acceptable for now, however it's incomplete in it's current form, which is why I'm writing this up. In this case we are using the Chain A clock to construct a timeout threshold that will eventually be validated on Chain B with the Chain B clock. In this case clock drift between Chain A and Chain B could cause the timeout threshold to be set incorrectly, however the result is less catastrophic. Worst case is we're creating some uncertainty as to whether or not the execution will occur or not on Chain B (not a runtime exception of the program). Ex ante inference about remote *evaluation conditions* is ok, so long as we set the evaluation conditions correctly. IBC provides a similar timeout facility, however IBC also has the notion of light client updates, which act as a heartbeat which commits to monotonically increasing timestamps from the remote chain. Client expiry then bounds the clock drift between chains, thereby preventing long-range attacks.

Takeaway is we're going to want something analogous to client updates and client expiration when we get to state proofs.

PR #219, opened 2025-02-14
Add execution time limit to subroutines

### LINKED ISSUE #137: Add an execution time limit to subroutines

keyleu:
Closes #137

Adds an expiration time to subroutines. This value determines how long a batch that is sent to a processor will be valid for execution from the time it is sent from the Authorization contract. It's an optional value but it's relevant for domains that use relayers without timeouts, like Hyperlane, as explained in #205.

How it works: the program owner will create his authorization and has the option to add an expiration_time to his subroutines, which is an absolute value in seconds. When messages are sent from this authorization, we will add this value to the current block time and forward it along with the rest of the batch info to the processor, this value will be the last valid timestamp that the messages can be executed. Everytime a batch is selected for execution (top from the queue in normal Processor and once it's received in the Lite Processor), we will check this value against the block time of that domain and execute it or not. 

This value can be 0 (very particular case) if the owner wants this to be immediately executed when sent (e.g. we send the messages and tick the processor on the same block on main domain), so we don't apply any specific validation on it, the program creator can decide for how long the messages for his subroutine are valid.

For this, we introduce a new execution result "Expired" so we can know if our subroutine expired or not. Since non-atomic batches might have been partially executed, we also provide the amount of execution functions along with the ExecutionResult.

It's implemented and tested for both Cosmwasm processor and EVM lite processor.


packages/authorization-utils/src/authorization.rs
```diff
@@ -143,18 +143,29 @@ impl Subroutine {
...
+    }
 }
 
 #[cw_serde]
 pub struct AtomicSubroutine {
     pub functions: Vec<AtomicFunction>,
     // Used for Atomic batches, if we don't specify retry logic then the functions won't be retried.
     pub retry_logic: Option<RetryLogic>,
+    // How long that a batch for this subroutine will be valid once it's sent from the authorization contract, in seconds.
+    pub expiration_time: Option<u64>,
```
bekauz, line 161: have we thought about enabling block-based expirations?

docs/src/authorizations_processors/processor.md
```diff
@@ -11,7 +11,8 @@ The Authorization contract will be the only address allowed to add message batch
 ### Execution
 
 When a processor is `Ticked`, the first `Message Batch` will be taken from the queue (`High` if there are batches there or `Med` if there aren’t).
-After taking them, we will execute them in different ways depending if the batch is `Atomic` or `NonAtomic`.
+After taking it, the first check the processor will do is check if the batch is expired. If that's the case, we will discard the batch and return an `Expired(executed_functions)` `ExecutionResult` to the Authorization contract. There might be a case that the batch is `NonAtomic` and it's already partially executed, therefore we return the number of functions that were executed before the expiration.
```
bekauz, line 14:
```suggestion
After taking it, the first check the processor will do is check if the batch is expired. If that's the case, the processor will discard the batch and return an `Expired(executed_functions)` `ExecutionResult` to the Authorization contract. There might be a case that the batch is `NonAtomic` and it's already partially executed, therefore we return the number of functions that were executed before the expiration.
```

docs/src/authorizations_processors/authorization_owner_actions.md
```diff
@@ -13,7 +13,7 @@
...
+  - Subroutine: set of functions in a specific order to be executed. Subroutines can be of two types: `Atomic` or `NonAtomic`. For the `Atomic` subroutines, we will provide an array of `Atomic` functions, an optional `expiration_time` and an optional `RetryLogic` for the entire subroutine. For the `NonAtomic` subroutines we will just provide an array of `NonAtomic` functions and an optional `expiration_time`. The `expiration_time` defines how long messages that are executing a subroutine will be valid for once they are sent from the authorization contract. This is particularly useful for domains that use relayers without timeouts (e.g. Hyperlane). If the `expiration_time` is not provided, the relayer can go down for an indefinite amount of time and the messages will still be valid and execute when it's back up. If the `expiration_time` is provided, the messages will be valid for that amount of time and if the relayer is down for longer than that, the messages will be considered expired once the execution is attempted in the Processor contract, returning an `Expired` result.
```
bekauz, line 16: maybe worth clarifying here about how the `expiration_time` actually gets applied to the messages

contracts/processor/src/contract.rs
```diff
@@ -193,6 +204,7 @@ fn enqueue_messages(
         msgs,
         subroutine,
         priority,
+        expiration_time,
```
bekauz, line 207: maybe worth validating this expiration time here? e.g. if it's already expired, error out right away?

bekauz (commented):
overall this looks good to me. just a few points regarding expiration time validation once the processor receives a batch and about block-based expiration configs.

docs/src/authorizations_processors/processor.md
uditvira, line 14:
```suggestion
After taking the `Message Batch`, the processor will first check if the batch is expired. If that's the case, the processor will discard the batch and return an `Expired(executed_functions)` `ExecutionResult` to the Authorization contract. There might be a case that the batch is `NonAtomic` and it's already partially executed, therefore the processor also returns the number of functions that were executed before the expiration.
```

docs/src/authorizations_processors/processor.md
```diff
@@ -11,7 +11,8 @@ The Authorization contract will be the only address allowed to add message batch
...
+If it's still valid, the processor will execute it in different ways depending if the batch is `Atomic` or `NonAtomic`.
```
uditvira, line 15:
```suggestion
If the batch has not expired, the processor will execute the batch according to whether it is `Atomic` or `NonAtomic`.
```

docs/src/authorizations_processors/authorization_owner_actions.md
keyleu: clarified

contracts/processor/src/contract.rs
keyleu: nice, added it. It's not an error but immediately send the callback with the expired status

packages/authorization-utils/src/authorization.rs
keyleu: not for now. we would need to query the block header of the opposite chain which can get complicated tbh. For now this solves our problems

contracts/authorization/src/contract.rs
```diff
@@ -749,12 +763,18 @@ fn retry_msgs(deps: DepsMut, env: Env, execution_id: u64) -> Result<Response, Co
                 callback_info.label.clone(),
                 &current_executions.checked_add(1).expect("Overflow"),
             )?;
+            // Calculate batch expiration time if Subroutine has one
+            let expiration_time = authorization.subroutine.get_expiration_time().map(|time| {
+                time.checked_add(env.block.time.seconds())
+                    .expect("Overflow")
```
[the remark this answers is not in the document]
keyleu: removed the expect. In the edge case that  the value is big enough that the checked_add returns None because it's too big, we won't pass expiration time and it can be executed any time, which makes sense to me

contracts/processor/src/contract.rs
keyleu: added an e2e test for this under polytone as this (sending the callback) will now be executed directly by the relayer instead of who ticks. Works as expected

contracts/authorization/src/contract.rs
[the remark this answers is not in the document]
bekauz: oh I didn't realize that `config.expiration_time` held a `u64` instead of `Uint64`, this makes sense

contracts/processor/src/contract.rs
bekauz, line 1: revisiting the processor it got me thinking whether it's worth to merge the `insert_messages` and `enqueue_messages` into a single function to reduce code? only difference is whether the messages get pushed to the end of the queue or at a custom index. the execute messages could stay the same, only diff would be that for `enqueue_messages` we would pass the end index. wdyt?

bekauz (approved):
lgtm, just one thought about the way we enqueue/insert messages into the queue.

contracts/processor/src/contract.rs
keyleu: yeah there is probably a lot of code reduction that could be done there, since its kinda audited code I don't want to touch it much but will open an issue for it to revisit

PR #220, opened 2025-02-14
Fix solidity CI

keyleu:
New foundry stable doesn't run optimizer by default so added optimizer flag.
Also fixed some tests that have a different behavior in this new version


bekauz (approved):
lgtm

PR #233, opened 2025-02-20
EVM: CCTPTransfer library

keyleu:
Adds a CCTPTransfer Library to transfer tokens using CCTP V1: simple library that will check that enough balance is available according to config (like we do in other transfer libraries like generic-ibc and neutron-ibc) and then submit the approve call to the ERC20 token and the depositForBurn call to the CCTP contract.

Tests are split in two: 
- standard tests that verify the config and balance checks
- script test that forks mainnet eth and tests against the real CCTP/USDC contracts

The encoder part is also set up so that we can use it in e2e testing.
Also added a doc section for the library that @uditvira can take a quick look at as well.


docs/src/libraries/evm/cctp_transfer.md
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  EVM --- DEST
+```
+
+## Functions
+
+| Function     | Parameters | Description                                                                                                   |
+| ------------ | ---------- | ------------------------------------------------------------------------------------------------------------- |
+| **Transfer** | -          | Transfer funds from the configured **input account** to the **mint recipient** on the **destination domain**. |
```
uditvira, line 39: I'm assuming you can, but to double-check: it is possible to add and constrain an authorization to update the config, correct?

keyleu: Yes, UpdateConfigs are added to the encoder and they are available for all libraries

docs/src/libraries/evm/cctp_transfer.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Configuration
+
+The library is configured on deployment using the `CCTPTransferConfig` type.
```
uditvira, line 43: Can we add a link to the [domain identifier table](https://developers.circle.com/stablecoins/supported-domains)?

docs/src/libraries/evm/cctp_transfer.md
keyleu: to extend on this, you can also update the processor and update the owner of a library, exactly like in CosmWasm.

docs/src/libraries/evm/cctp_transfer.md
keyleu: added

solidity/script/CCTPTransfer.script.sol
bekauz, line 1: probably not relevant for now but in the future would be nice to remove the consts and make those fields configurable to support non-Ethereum evms

solidity/src/libraries/CCTPTransfer.sol
```diff
@@ -0,0 +1,125 @@
...
+     * @dev Configuration struct for token transfer parameters.
+     * @param amountToTransfer The number of tokens to transfer. If set to 0, the entire balance is transferred.
+     * @param mintRecipient The recipient address (in bytes32 format) on the destination chain where tokens will be minted.
+     * @param inputAccount The account from which tokens will be debited.
+     * @param destinationDomain The domain identifier for the destination chain.
+     * @param cctpTokenMessenger The CCTP Token Messenger contract.
+     * @param transferToken The ERC20 token address that will be transferred.
+     */
+    struct CCTPTransferConfig {
+        uint256 amountToTransfer; // If we want to transfer all tokens, we can set this to 0.
```
bekauz, line 25: mentioned this on cw side but imho `amount` is clear enough here

solidity/src/libraries/CCTPTransfer.sol
```diff
@@ -0,0 +1,125 @@
...
+     */
+    function transfer() external onlyProcessor {
+        // Retrieve the current configuration into a local variable.
+        CCTPTransferConfig memory _config = config;
+
+        // Check the token balance of the input account.
+        uint256 balance = IERC20(_config.transferToken).balanceOf(address(_config.inputAccount));
+        if (balance == 0 || balance < _config.amountToTransfer) {
+            revert("Insufficient balance");
+        }
```
bekauz, line 104: thinking about a case where `_config.amountToTransfer = 0` and `balance = 0`. rn it would return insufficient balance error which feels a bit vague.  maybe worth to split up these checks into two for clearer errors? also would be nice to hit both of these cases with unit tests

solidity/src/libraries/CCTPTransfer.sol
```diff
@@ -0,0 +1,125 @@
...
+
+        // Determine the amount to transfer:
+        // If amountToTransfer is greater than 0, use that value; otherwise, transfer the full balance.
+        uint256 _amountToTransfer = _config.amountToTransfer > 0 ? _config.amountToTransfer : balance;
+
+        // Encode the approval call: this allows the CCTP Token Messenger to spend the tokens.
+        bytes memory encodedApproveCall =
```
bekauz, line 111: _I'm not sure how approval calls work in cases where the approval is already granted so this may not make much sense, but..._

on a first sight doing approvals on every transfer call feels a little excessive, especially when thinking about long-running programs like Vaults (which will be constantly pinging this transfer call). wyt if we split away the approval into a separate function or even the constructor? that way transfer would become more lean and over the long run save some gas

bekauz: following on this, what happens if some account grants approval to the CCTP Token Messenger via this function call and later down the line that account wants to revoke the approval? would it be possible to do in a direct call from the account to the token messenger (bypassing this library)? just wondering if doing approval via intermediate contract could impact something

bekauz (commented):
overall looks pretty good! will revisit tomorrow

contracts/encoders/evm-encoder/src/libraries/cctp_transfer.rs
```diff
@@ -0,0 +1,87 @@
...
+// We need to define a config and functions for this library as we don't have a CosmWasm equivalent
+#[cw_serde]
+/// Struct representing the library configuration.
+pub struct LibraryConfig {
+    /// The input address for the library.
+    pub input_addr: LibraryAccountType,
+    /// The mint recipient for the library. Bytes32 representation of the address in solidity.
+    pub mint_recipient: Binary,
+    /// Amount to transfer
+    pub amount_to_transfer: Uint128,
```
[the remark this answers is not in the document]
keyleu: changed!

contracts/encoders/evm-encoder/src/libraries/cctp_transfer.rs
```diff
@@ -0,0 +1,87 @@
...
+
+// We need to define a config and functions for this library as we don't have a CosmWasm equivalent
+#[cw_serde]
+/// Struct representing the library configuration.
+pub struct LibraryConfig {
+    /// The input address for the library.
+    pub input_addr: LibraryAccountType,
+    /// The mint recipient for the library. Bytes32 representation of the address in solidity.
+    pub mint_recipient: Binary,
+    /// Amount to transfer
```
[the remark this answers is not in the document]
keyleu: changed!

contracts/encoders/evm-encoder/src/libraries/cctp_transfer.rs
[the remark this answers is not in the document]
keyleu: changed to Uint256! Initially was just using Uint128 because I could convert from native u128

solidity/script/CCTPTransfer.script.sol
keyleu: Yeah we just write different scripts in that case or pass the addresses as arguments, for now I'll just keep it like this

solidity/src/libraries/CCTPTransfer.sol
keyleu: changed!

solidity/src/libraries/CCTPTransfer.sol
keyleu: Split it into two errors and extended it in the unit tests

solidity/src/libraries/CCTPTransfer.sol
keyleu: So basically how approval works: we approve the transfer of those tokens, and the approval is decreased with that amount. Therefore we need to approve each time we transfer (similar to Cw20) because each time we transfer we "consume" the approval. It's not like its approved forever.

Example:

1) I approve the library to transfer 20 tokens.
2) CCTP Transfer library transfers the tokens to itself and burns them, consuming the approval of 20 tokens.
3) Current approval is now 0, if we try to transfer without approving now it would fail because we have no approval.

Regarding your second question. There will never be anything to revoke because we are always consuming the approval we are granting and we are not leaving any "approval" hanging there. 

The gas used for this is actually little as we are getting refunded as well when we consume and delete the state. It's how ERC20 tokens are transfered on Ethereum all the time. I don't think it's worth it to approve in advance andn then here do a query instead because we want to get the gas refund of deleting it form the state when executing.

Additionally, this library will be used with big amounts, CCTP has fees that makes it pointless to use with small amounts, if we compare that with the gas it's kinda insignificant

bekauz: thanks for this, it explains a lot. still curious if in the future we can implement some kind of pre-approval flow for recurring situations like vault deposit forwarding from one place to another. but it's an optimization so no need to worry about that now.

bekauz (approved):

contracts/encoders/evm-encoder/src/libraries/cctp_transfer.rs
```diff
@@ -0,0 +1,87 @@
...
+use super::{get_update_ownership_call, get_update_processor_call};
+
+// We need to define a config and functions for this library as we don't have a CosmWasm equivalent
+#[cw_serde]
+/// Struct representing the library configuration.
+pub struct LibraryConfig {
+    /// The input address for the library.
+    pub input_addr: LibraryAccountType,
+    /// The mint recipient for the library. Bytes32 representation of the address in solidity.
+    pub mint_recipient: Binary,
```
Art3miX, line 20: Shouldn't we avoid the underlying protocol specifics? 
`mint_recipient` is CCTP specifics, in IBC transfer you also burn and mint, but IBC language doesn't include it.

I think we should try to keep a uniformed language between libraries that perform the same action, I think `recipient_addr` would be better name here and in any future transfer libraries.

solidity/src/libraries/interfaces/cctp/ITokenMessenger.sol
Art3miX, line 1: Why do we have it? can't we import the CCTP one?

solidity/test/libraries/CCTPTransfer.t.sol
```diff
@@ -0,0 +1,160 @@
...
+            destinationDomain: destinationDomain,
+            cctpTokenMessenger: ITokenMessenger(address(stubTokenMessenger)),
+            transferToken: address(token)
+        });
+        bytes memory configBytes = abi.encode(validConfig);
+        vm.prank(owner);
+        cctpTransfer = new CCTPTransfer(owner, processor, configBytes);
+
+        vm.prank(processor);
+        vm.expectRevert("Nothing to transfer");
```
Art3miX, line 126: out of interest, why not use errors? they are more gas efficient

solidity/src/libraries/interfaces/cctp/ITokenMessenger.sol
keyleu: there's no package for it and importing CCTP repo doesn't work for 2 reasons:
- We need to set it up because they are using git submodules (would need extra set up but could be done)
- Contracts are using an old compiler version which makes it incompatible with contracts that are not using the same version.

Checked how other integrators are doing it like Skip and wormhole and they do the same.

solidity/test/libraries/CCTPTransfer.t.sol
keyleu: Because we send errors back to authorization contract in the execution result, if we are sending errors instead of strings we will be getting some abi_encoded bytes that we don't know how to decode them into unless we look into the contract types and will be a nightmare to debug for users.
Strings they can always convert the bytes to string and read what went wrong

contracts/encoders/evm-encoder/src/libraries/cctp_transfer.rs
Art3miX: Adding to this, we should use `LibraryAccountType` for all addresses in our libraries, working on another PR that fixes it for the IBC transfer.
The reason is that recipient will probably be an account, and if we want the manager to replace addresses in the config, we need to use `LibraryAccountType` type to associate an account_id to instantiated address.
In this case me might add another variant to the enum which is not a string but Binary, or not if you can work with the string.

keyleu: This is for the config, I think it will be much easier for the program creator/owner to have the same/similar API as the external protocol we are integrating so they can easily go through the docs of that bridge and know what these fields are and avoid confusions.

We can change it though, I don't see an issue here, just thinking what could be best to avoid confusions. 
When users send messages they will just send the transfer(), they won't see these fields

PR #244, opened 2025-02-24
Valence ICA

keyleu:
Adds Valence ICA.

This account will be created on Neutron with a RemoteDomainInfo and libraries can use it to trigger remote execution of protobuf messages on a remote domain via IBC.
Used covenants as reference for implementation. One Valence ICA will work with a single Cosmos ICA.

Testing will be done once in my next PR, once I have the first library (CCTP library) that will be used to interact with a Valence ICA and execute a proto message on Noble.

Added @uditvira to review docs update, which contains a refactor of the account section since we now have 3 different account types.


contracts/accounts/interchain_account/src/state.rs
```diff
@@ -0,0 +1,11 @@
+use cosmwasm_std::{Addr, Empty};
+use cw_storage_plus::{Item, Map};
+
+use crate::msg::{IcaState, RemoteDomainInfo};
+
+// Approved libraries that can execute actions on behalf of the account
+pub const APPROVED_LIBRARIES: Map<Addr, Empty> = Map::new("libraries");
```
Art3miX, line 7: We should not use Empty in storage, from Simon:
"Don't use Empty in storage. Empty needs to disappear at some point (probably CosmWasm 3.0)"

The better way would be to use 1/0 as u8, it is better then bool, because bool is parsed into a string "true" "false".

docs/src/examples/crosschain_vaults.md
```diff
@@ -84,7 +84,7 @@ We'll need the following Libraries on Neutron:
...
 - **Bridge Transfer**: To transfer funds from the Neutron Withdraw Account to the Ethereum Withdraw Account.
 
-Note that the Accounts mentioned here are the standard [Valence Base Accounts](../components/base_accounts.md). Th Bridge Transfer library will depend on the token being transferred, but will offer similar functionality to the [IBC Transfer](../libraries/cosmwasm/generic_ibc_transfer.md) library. The Position Depositor and Withdrawer will depend on the type of position, but can be similar to the [Liqudity Provider](../libraries/cosmwasm/astroport_lper.md) and [Liquidity Withdrawer](../libraries/cosmwasm/astroport_withdrawer.md).
+Note that the Accounts mentioned here are the standard [Valence Base Accounts](../accounts/base_accounts.md). Th Bridge Transfer library will depend on the token being transferred, but will offer similar functionality to the [IBC Transfer](../libraries/cosmwasm/generic_ibc_transfer.md) library. The Position Depositor and Withdrawer will depend on the type of position, but can be similar to the [Liqudity Provider](../libraries/cosmwasm/astroport_lper.md) and [Liquidity Withdrawer](../libraries/cosmwasm/astroport_withdrawer.md).
```
Art3miX, line 87:
```suggestion
Note that the Accounts mentioned here are the standard [Valence Base Accounts](../accounts/base_accounts.md). The Bridge Transfer library will depend on the token being transferred, but will offer similar functionality to the [IBC Transfer](../libraries/cosmwasm/generic_ibc_transfer.md) library. The Position Depositor and Withdrawer will depend on the type of position, but can be similar to the [Liqudity Provider](../libraries/cosmwasm/astroport_lper.md) and [Liquidity Withdrawer](../libraries/cosmwasm/astroport_withdrawer.md).
```

Art3miX (changes requested):

contracts/accounts/interchain_account/src/contract.rs
```diff
@@ -0,0 +1,293 @@
...
+        ensure!(
+            ica_registration_fee.iter().all(|coin| {
+                deps.querier
+                    .query_balance(&env.contract.address, &coin.denom)
+                    .unwrap_or_default()
+                    .amount
+                    >= coin.amount
+            }),
+            ContractError::NotEnoughBalanceForIcaRegistration
+        );
```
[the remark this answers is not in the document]
keyleu: went through docs about this and there's no details, just that the registration needs to pay the fees, which have been an array of 1 element (10 NTRN )since the module was created. The fact is that the query returns an array of coins and this is how it's currently done in the covenants, pinging @bekauz here as he might know more.

contracts/accounts/interchain_account/src/state.rs
keyleu: We can change all those when/if the time comes but rather stay consistent with how we are doing it everywhere now. We can change it everywhere in a separate PR and we would be saving some gas during storage because we would write less bytes that way.

Art3miX: Less about gas and more about deprecation / right way of doing it.

contracts/accounts/interchain_account/src/contract.rs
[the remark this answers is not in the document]
Art3miX: My concern here is that I think they wanted to allow multiple denoms to pay the fee, but we are not checking for just one, rather check that we have balance for all of them.

contracts/accounts/interchain_account/src/state.rs
keyleu: yeah, we can change it once it's deprecated

contracts/accounts/interchain_account/src/contract.rs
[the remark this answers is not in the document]
keyleu: actually went through Neutron code and
https://github.com/neutron-org/neutron/blob/d9e23d27d00353e53610ea83ecc81e2f9452272f/x/interchaintxs/keeper/keeper.go#L79C1-L81C3

So it will be fine as long as we send one of the fees. Changed to check that we at least have 1 coin that matches the condition

Art3miX (approved):

contracts/accounts/interchain_account/src/contract.rs
```diff
@@ -0,0 +1,293 @@
...
+    msg: InstantiateMsg,
+) -> StdResult<Response> {
+    set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
+    cw_ownable::initialize_owner(deps.storage, deps.api, Some(&msg.admin))?;
+
+    msg.approved_libraries.iter().try_for_each(|library| {
+        APPROVED_LIBRARIES.save(deps.storage, deps.api.addr_validate(library)?, &Empty {})
+    })?;
+
+    REMOTE_DOMAIN_INFO.save(deps.storage, &msg.remote_domain_information)?;
```
bekauz, line 37: should we do some simple validation on the remote domain config?

contracts/accounts/interchain_account/src/contract.rs
```diff
@@ -0,0 +1,293 @@
...
+            .add_attribute("library", library_addr))
+    }
+
+    pub fn try_register_ica(
+        deps: DepsMut<NeutronQuery>,
+        env: Env,
+    ) -> Result<Response, ContractError> {
+        // First we verify that we are in the correct state to allow ICA creation
+        let state = ICA_STATE.load(deps.storage)?;
+        if state.ne(&IcaState::NotCreated) || state.ne(&IcaState::Closed) {
```
bekauz, line 116: it's a bit hard to see this way what are the possible states, maybe a match would be nicer?

contracts/accounts/interchain_account/src/contract.rs
```diff
@@ -0,0 +1,293 @@
...
+    }
+}
+
+#[cfg_attr(not(feature = "library"), entry_point)]
+pub fn sudo(deps: DepsMut, _env: Env, msg: SudoMsg) -> StdResult<Response> {
+    match msg {
+        SudoMsg::Response { request, data } => {
+            // If the channel closed, we need to update the state to Closed to allow recreation
+            if request.sequence.is_none() || request.source_channel.is_none() {
+                ICA_STATE.save(deps.storage, &IcaState::Closed)?;
```
bekauz, line 242: is there any practical difference between `IcaState::Closed` and `IcaState::NotCreated`?

contracts/accounts/interchain_account/src/msg.rs
```diff
@@ -0,0 +1,64 @@
...
+#[cw_serde]
+pub struct RemoteDomainInfo {
+    pub connection_id: String,
+    pub ica_timeout: Uint64,
+}
+
+#[cw_serde]
+pub enum IcaState {
+    NotCreated, // Not created yet
+    Closed,     // Was created but closed, so creation should be retriggered
```
bekauz, line 43: mentioned in the contract but `Closed` variant feels unnecessary, wyt?

packages/account-utils/src/ica.rs
```diff
@@ -0,0 +1,109 @@
...
+}
+
+#[cw_serde]
+pub struct Params {
+    pub msg_submit_tx_max_messages: Uint64,
+    pub register_fee: Vec<Coin>,
+}
+
+#[cw_serde]
+pub struct QueryParamsResponse {
```
bekauz, line 46: does importing this from neutron cause serde issues?

packages/account-utils/src/ica.rs
```diff
@@ -0,0 +1,109 @@
...
+
+#[cw_serde]
+pub struct OpenAckVersion {
+    pub version: String,
+    pub controller_connection_id: String,
+    pub host_connection_id: String,
+    pub address: String,
+    pub encoding: String,
+    pub tx_type: String,
+}
```
bekauz, line 21: I remember defining this during my work on the ic querier, maybe we can move this and some other ica/icq/neutron-ibc related utils into some common package for reuse?

packages/account-utils/src/ica.rs
```diff
@@ -0,0 +1,109 @@
...
+        interchain_account_id,
+        register_fee,
+    };
+
+    #[allow(deprecated)]
+    CosmosMsg::Stargate {
+        type_url: "/neutron.interchaintxs.v1.MsgRegisterInterchainAccount".to_string(),
+        value: Binary::from(msg_register_interchain_account.to_bytes().unwrap()),
+    }
+}
```
bekauz, line 74: why not use the `NeutronMsg::register_interchain_account` helper? would avoid stargate

packages/account-utils/src/ica.rs
```diff
@@ -0,0 +1,109 @@
...
+        timeout,
+        fee: Some(fee),
+    };
+
+    #[allow(deprecated)]
+    CosmosMsg::Stargate {
+        type_url: "/neutron.interchaintxs.v1.MsgSubmitTx".to_string(),
+        value: Binary::from(msg_submit_tx.to_bytes().unwrap()),
+    }
+}
```
bekauz, line 109: same here, can use `NeutronMsg::submit_tx` helper

contracts/accounts/interchain_account/src/msg.rs
keyleu: I just want it for informational purposes, so that we can query and see if it closed for some reason or it just hasn't been created yet

packages/account-utils/src/ica.rs
keyleu: I don't to use NeutronMsgs, also it goes through bindings which I want to avoid seeing how they mess around with bindings often.

packages/account-utils/src/ica.rs
keyleu: kinda prefer using protos directly, I want to avoid NeutronMsgs and go through wasm bindings which are deprecated

contracts/accounts/interchain_account/src/contract.rs
keyleu: added!

contracts/accounts/interchain_account/src/contract.rs
keyleu: changed for a match

packages/account-utils/src/ica.rs
keyleu: right, didn't see that. Moved everything under neutron.rs in valence-ibc-utils and import it with the neutron feature for both icq and ica

packages/account-utils/src/ica.rs
keyleu: yup, it's not serializable there. Neutron-std has it but rn it's not compatible with our testing framework, we have a TODO in the Cargo.toml to upgrade and then we can get rid of all deprecated Stargate messages, use neutron-std instead of neutron-sdk, and take all these structs from neutron-std directly.
When we have a proper testing framework where we can test with latest Neutron version we can remove all this.

contracts/accounts/interchain_account/src/contract.rs
```diff
@@ -0,0 +1,299 @@
...
+
+        let library_addr = deps.api.addr_validate(&library)?;
+        APPROVED_LIBRARIES.remove(deps.storage, library_addr.clone());
+
+        Ok(Response::new()
+            .add_attribute("method", "remove_library")
+            .add_attribute("library", library_addr))
+    }
+
+    pub fn try_register_ica(
```
uditvira, line 111: Does Neutron support ICAs over unordered channels? If so, we much prefer to use that so that we don't fall into a state where the channels close during timeouts.

packages/ibc-utils/src/neutron.rs
```diff
@@ -183,3 +187,109 @@ fn get_transfer_fee(ibc_fee: IbcFee) -> neutron_sdk::proto_types::neutron::feere
...
+        })
+        .collect();
+
+    let msg_submit_tx = MsgSubmitTx {
+        from_address: sender.to_string(),
+        interchain_account_id,
+        connection_id,
+        msgs: any_msgs,
+        memo,
+        timeout,
```
uditvira, line 286: Do we have tests for timeouts? I.e., if a message times out, does the channel recovery work as expected?

contracts/accounts/interchain_account/src/contract.rs
keyleu: Not sure? Didn't find anything in the docs. In any case they only have v1 of interchaintxs module and a single entrypoint to register the ICA https://github.com/neutron-org/neutron/blob/e74810731c72f200fbeb4ca4a07c16ab9f971382/proto/neutron/interchaintxs/v1/tx.proto#L22 so we only have this way to register an ICA

packages/ibc-utils/src/neutron.rs
keyleu: Tests are not included in this PR, I'm planning on adding our first ICA library in the next PR and there do all the ICA integration tests together with the ICA library, including the timeout and successful recovery

contracts/accounts/interchain_account/src/contract.rs
keyleu: going through docs looks like they are using ordered channels, asked in the group to confirm

contracts/accounts/interchain_account/src/contract.rs
keyleu: keeping it, reason below

docs/src/accounts/interchain_accounts.md
```diff
@@ -0,0 +1,97 @@
...
+In this message, the `connection_id` of the remote domain and the timeout for the ICA messages are specified.
+
+### Execute Methods
+
+```rust
+pub enum ExecuteMsg {
+    ApproveLibrary { library: String }, // Add library to approved list (only admin)
+    RemoveLibrary { library: String },  // Remove library from approved list (only admin)
+    ExecuteIcaMsg { msgs: Vec<ProtobufAny> }, // Execute a protobuf message on the ICA
+    RegisterIca {},                     // Register the ICA on the remote chain
```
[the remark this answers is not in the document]
uditvira: That works.

bekauz (approved):
lgtm

PR #249, opened 2025-02-27
Drop Liquid unstaker

keyleu:
Drop liquid unstaker library.

Has two flows:
- LiquidUnstake, similar to LiquidStake but instead sends the LS asset and gets an NFT voucher that will be used later to claim. This NFT voucher is sent back to the account that unstaked.

- Claim. Sends the NFT voucher to the withdrawal manager contract with the hook message to trigger the withdraw. This will claim (if claimable) and deposit into output account.

Bumped optimizer as cw-nfts is using rust 1.81 and we were still using an old optimizer that was optimizing for 1.78. 

Leaving all validations to the corresponding drop contracts because the storage is not stable across version and it would break contracts if they upgrade their mainnet contracts. Validations in the drop contracts provide meaningful errors.


contracts/libraries/drop-liquid-unstaker/README.md
```diff
@@ -0,0 +1,61 @@
...
+    pub input_addr: LibraryAccountType,
+    pub output_addr: LibraryAccountType,
+    // Address of the liquid unstaker contract (drop core contract)
+    pub liquid_unstaker_addr: String,
+    // Address of the withdrawal_manager_addr (drop withdrawal manager)
+    pub withdrawal_manager_addr: String,
+    // Address of the voucher NFT contract that we get after unstaking and we use for the claim
+    pub voucher_addr: String,
+    // Denom of the asset we are going to liquid unstake
+    pub denom: String,
```
Art3miX, line 59: Just a note, the library config is what users will need to write, I understand that in this case we want to get this info from outside because of some instability, but that means the users will need to manually query and get this info somehow.

contracts/libraries/drop-liquid-unstaker/src/contract.rs
```diff
@@ -0,0 +1,205 @@
...
+                // Wrap the unstake msg to be executed on behalf of the input account
+                let input_account_msgs = execute_on_behalf_of(vec![unstake_msg], &cfg.input_addr)?;
+
+                Ok(Response::new()
+                    .add_message(input_account_msgs)
+                    .add_attribute("method", "liquid_unstake"))
+            }
+            FunctionMsgs::Claim { token_id } => {
+                // Verify that voucher belongs to input account
+                verify_voucher_ownership(
```
Art3miX, line 90: Does it mint a single NFT per user? or a user can have multiple NFTs per multiple stakes? 
The reason I ask is that you need to know the token id here, this is a bit annoying in building the authorization or calling it, if only single token id is possible, you can query if the input account holds any NFT and just get that.

Of course if multiple NFTs are possible, the caller will have to know the correct id to use.

Art3miX (approved):

contracts/libraries/drop-liquid-unstaker/README.md
keyleu: Yes the program owner needs to set this up correctly. This information is all available in the drop docs so no need to query, it's all there.

contracts/libraries/drop-liquid-unstaker/src/contract.rs
keyleu: Multiple NFTs are possible. We are allowing as many unstakes as they want and they need to provide the ID they want to claim for each claim

uditvira (approved):
Docs are perfect, thanks!

contracts/libraries/drop-liquid-unstaker/src/contract.rs
```diff
@@ -0,0 +1,205 @@
...
+            .query_wasm_smart(
+                voucher_contract.clone(),
+                &cw721_base::msg::QueryMsg::OwnerOf {
+                    token_id: token_id.clone(),
+                    include_expired: None,
+                },
+            )
+            .map_err(|_| LibraryError::ExecutionError("Voucher does not exist".to_string()))?;
+
+        ensure_eq!(
```
bekauz, line 133: TIL there's `ensure_eq` and `ensure_ne` and not just `ensure`, cool

contracts/libraries/drop-liquid-unstaker/README.md
```diff
@@ -0,0 +1,61 @@
+# Valence Drop Liquid Unstaker library
+
+The **Valence Drop Liquid Unstaker** library allows to **liquid unstake** an asset from an **input account** from the [Drop protocol](https://docs.drop.money/) and to **withdraw** the claim once it's withdrawable and deposit the asset into the **output account**. It is typically used as part of a **Valence Program**. In that context, a **Processor** contract will be the main contract interacting with the Forwarder library.
```
bekauz, line 3: feels like this can be phrased better, something along the lines of redeeming the liquid staking token for the underlying assets. @uditvira wyt?

bekauz (approved):
lgtm. only feedback would be around the phrasing and function naming. not sure if we want to follow the terminology used by drop (bond/unbond) or have our own that would apply to all LS providers.

docs/src/libraries/cosmwasm/drop_liquid_unstaker.md
```diff
@@ -0,0 +1,68 @@
+# Valence Drop Liquid Unstaker library
+
+The **Valence Drop Liquid Unstaker** library allows to **liquid unstake** an asset from an **input account** from the [Drop protocol](https://docs.drop.money/) and to **withdraw** the claim once it's withdrawable and deposit the asset into the **output account**. It is typically used as part of a **Valence Program**. In that context, a **Processor** contract will be the main contract interacting with the Forwarder library.
```
uditvira, line 3:
```suggestion
The **Valence Drop Liquid Unstaker** library allows liquid staked tokens (e.g., dNTRN or dATOM) to be redeem for underlying assets (e.g., NTRN or ATOM) through the [Drop protocol](https://docs.drop.money/). The liquid staked asset must be available in the **input account**. When the library's function to redeem the staked assets (`LiquidUnstake`) is invoked, the library issues a withdraw request to the Drop protocol generating a tokenized voucher that is held by the input account. This tokenized voucher can be used to claim the underlying assets (represented as an NFT). Note that the underlying assets are not withdrawn immediately, as the Drop protocol unstakes assets asynchronously. At a later time, when the underlying assets are available for withdraw, the library's claim function can be invoked with the voucher as an argument. This function will wtihdraw the underlying assets and deposit them into the **output account**. 

This library is typically used as part of a **Valence Program**. In that context, a **Processor** contract will be the main contract interacting with the Forwarder library.
```

contracts/libraries/drop-liquid-unstaker/README.md
```diff
@@ -0,0 +1,61 @@
+# Valence Drop Liquid Unstaker library
+
+The **Valence Drop Liquid Unstaker** library allows liquid staked tokens (e.g., dNTRN or dATOM) to be redeem for underlying assets (e.g., NTRN or ATOM) through the [Drop protocol](https://docs.drop.money/). The liquid staked asset must be available in the **input account**. When the library's function to redeem the staked assets (`LiquidUnstake`) is invoked, the library issues a withdraw request to the Drop protocol generating a tokenized voucher that is held by the input account. This tokenized voucher can be used to claim the underlying assets (represented as an NFT). Note that the underlying assets are not withdrawn immediately, as the Drop protocol unstakes assets asynchronously. At a later time, when the underlying assets are available for withdraw, the library's claim function can be invoked with the voucher as an argument. This function will withdraw the underlying assets and deposit them into the **output account**.
```
bekauz, line 3:
```suggestion
The **Valence Drop Liquid Unstaker** library allows liquid staked tokens (e.g., dNTRN or dATOM) to be redeemed for underlying assets (e.g., NTRN or ATOM) through the [Drop protocol](https://docs.drop.money/). The liquid staked asset must be available in the **input account**. When the library's function to redeem the staked assets (`LiquidUnstake`) is invoked, the library issues a withdraw request to the Drop protocol generating a tokenized voucher that is held by the input account. This tokenized voucher can be used to claim the underlying assets (represented as an NFT). Note that the underlying assets are not withdrawn immediately, as the Drop protocol unstakes assets asynchronously. At a later time, when the underlying assets are available for withdrawal, the library's claim function can be invoked with the voucher as an argument. This function will withdraw the underlying assets and deposit them into the **output account**.
```

Cargo.toml
```diff
@@ -46,6 +46,8 @@ cw-ownable      = "2.0.0"
 cw-utils        = "2.0.0"
 cw-storage-plus = "2.0.0"
 cw2             = "2.0.0"
+cw721           = { git = "https://github.com/public-awesome/cw-nfts", rev = "99dbf7e" }
+cw721-base      = { git = "https://github.com/public-awesome/cw-nfts", rev = "99dbf7e" }
```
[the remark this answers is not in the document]
keyleu: their last release, 0.18.0 is not compatible because it's using an old cosmwasm version, they haven't done a release lately. I'm going to try to contact one of the owners to see if they can do a release because they haven't done one in a while and then we can change this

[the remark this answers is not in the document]
keyleu: pinged them on TG, if they do I'll make another PR to update this, will keep an eye on it

PR #251, opened 2025-02-28
Add noble local-ic config setup

keyleu:
Adding config here to spin-up noble to set up the chain with all the initial values required to test bridging assets. Doing this PR first in case @bekauz you need to spin up noble for some reason, which I don't think you'll be doing soon but anyways.

Additional things that need to be done to be able to trigger the bridging (reproduced in CLI) and taken from their go integration tests in noble-assets/noble:

```bash
# Get the module account for cctp module which happens to be noble12l2w4ugfz4m6dd73yysz477jszqnfughxvkss5
nobled q auth module-accounts

# Mint some USDC to myself
nobled tx fiat-tokenfactory mint noble1hj5fveer5cjtn4wd6wstzugjfdxzl0xpw0865d 1000000uusdc --from admin --chain-id localnoble-1 -y

# Configure the cctp module account to allow minting
nobled tx fiat-tokenfactory configure-minter-controller noble1hj5fveer5cjtn4wd6wstzugjfdxzl0xpw0865d noble12l2w4ugfz4m6dd73yysz477jszqnfughxvkss5 --from admin --chain-id localnoble-1 -y

nobled tx fiat-tokenfactory configure-minter noble12l2w4ugfz4m6dd73yysz477jszqnfughxvkss5 10000000000uusdc --from admin --chain-id localnoble-1 -y

# Add a remote token messenger for a destination domain (in this case we are using domain-id 0), my guess is this is the contract address that mints
nobled tx cctp add-remote-token-messenger 0 0xfce4ce85e1f74c01e0ecccd8bbc4606f83d3fc90 --from admin --chain-id localnoble-1 -y

# Link the token, uusdc representation on that domain
nobled tx cctp link-token-pair uusdc 0x07865c6E87B9F70255377e024ace6630C1Eaa37F 0 --from admin --chain-id localnoble-1 -y

# Call the deposit-for burn (bridging). Here im bridging 1000 uusdc to domain 0 and recipient address is 0xfce... (could be any address)
nobled tx cctp deposit-for-burn 1000 0 0xfce4ce85e1f74c01e0ecccd8bbc4606f83d3fc90 uusdc --from admin --chain-id localnoble-1 -y
```

Next steps when I'm writing my tests will be writing a helper that does all this via rust before writing my tests

### DIFF
--- a/e2e/chains/neutron_noble.json
+++ b/e2e/chains/neutron_noble.json
@@ -1,0 +1,277 @@
+{
+  "chains": [
+    {
+      "name": "gaia",
+      "chain_id": "localcosmos-1",
+      "denom": "uatom",
+      "binary": "gaiad",
+      "bech32_prefix": "cosmos",
+      "docker_image": {
+        "version": "v15.0.0-rc2"
+      },
+      "gas_prices": "0%DENOM%",
+      "chain_type": "cosmos",
+      "coin_type": 118,
+      "trusting_period": "336h",
+      "gas_adjustment": 1.3,
+      "number_vals": 1,
+      "number_node": 0,
+      "ibc_paths": ["gaia-noble"],
+      "debugging": true,
+      "block_time": "1s",
+      "genesis": {
+        "modify": [
+          {
+            "key": "app_state.gov.params.voting_period",
+            "value": "3s"
+          },
+          {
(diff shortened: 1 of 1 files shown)


bekauz (approved):
lgtm! maybe the setup steps you mentioned can be done using the chain-client-utils :)

PR #252, opened 2025-03-01
strategist domain clients

### LINKED ISSUE #245: strategist domain clients
# Description

The orchestrator will need a unified interface for interacting with different domains in order to build strategies in an efficient manner.

## Shared client behavior

Shared client behavior can be achieved by defining a common trait, `StrategistClient`, which would be implemented by structs that identify a particular chain (e.g. `Ethereum` or `Neutron`).

Method signatures that this trait should define need to cover following:

- Querying account balances
- Querying contract state
- Executing transactions

## Chain Clients

Chain clients will be the points of interaction by the strategist orchestrator. Clients should be instantiated with any parameters needed in order to be able to sign and execute transactions. This involves:

- RPC url
- seed phrase for the signer  

Clients should interact with their respective domains via existing libraries:

- cosmos-sdk should leverage `cosmrs`
- evm should leverage `alloy-rs`

bekauz:
closes #245 

# Description

Base implementation for strategist domain client utils to enable off-chain interactions with chain nodes.

To see these utils in action, look at the unit tests of `ethereum.rs` and `neutron.rs` (they expect local-ic running on the specified ports). After a specific client implementation is instantiated with the relevant chain/signer identifying information, clients are ready to perform the following functionality in an async manner:

- query block headers
- query balances
- execute smart contracts
- query smart contracts
- ibc transfer (cosmos-only)

Defined clients can be used standalone or inside local-ic.

## Design

 To minimize the overhead of enabling a new chain, the design relies heavily on traits that provide default implementations.

### Cosmos

Cosmos clients have two main traits available for them: `BaseClient` and `WasmClient`.
Chains that do not have wasmd module can ignore `WasmClient` and just implement `BaseClient`. Chains that want both can implement both, which looks like this:

```rust
// OsmosisClient defined above

/// osmosis is a base cosmos chain
#[async_trait]
impl BaseClient for OsmosisClient {}

/// osmosis has a wasm module
#[async_trait]
impl WasmClient for OsmosisClient {}
```

Default trait implementations can be overriden. For example, Neutron has a custom ibc module where ibc transfers function differently. It still wants to inherit vast majority of the default `BaseClient` functionality, so it can override just the transfer method as follows:

```rust
// NeutronClient defined above

#[async_trait]
impl BaseClient for NeutronClient {
    /// neutron has custom ibc logic so we override the default BaseClient ibc_transfer
    async fn ibc_transfer(args) -> Result<TransactionResponse, StrategistError> {
         // custom logic
    }
}

#[async_trait]
impl WasmClient for NeutronClient {}
```

Both `BaseClient` and `WasmClient` depend on `GrpcSigningClient` trait which describes how the client is meant to sign transactions and open a grpc channel.

### EVM

EVM clients have just one trait for now - `EvmBaseClient`, which depends on `RequestProviderClient`.
These traits serve similar purpose to the ones described in cosmos section above, although they rely on different args and response types so I didn't bother normalizing them (can be done later).


packages/chain-client-utils/Cargo.toml
```diff
@@ -0,0 +1,45 @@
...
+edition    = { workspace = true }
+license    = { workspace = true }
+version    = { workspace = true }
+repository = { workspace = true }
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+# generic imports
+localic-utils = { workspace = true }
```
keyleu, line 12: don't think we should add localic-utils dependency just to get chain denom and chain prefix. Probably worth defining them in each chain's client. Feels a bit weird having this here

packages/chain-client-utils/Cargo.toml
```diff
@@ -0,0 +1,45 @@
...
+valence-e2e   = { workspace = true }
```
keyleu, line 13: this one is only being used for tests right? can we add it as a dev dependency? Should probably add
#[ignore = "internal test"] to the tests so that CI doesn't try to run them?

packages/chain-client-utils/src/cosmos/wasm_client.rs
```diff
@@ -0,0 +1,81 @@
...
+        let wasm_tx = MsgExecuteContract {
+            sender: signing_client.address.clone(),
+            contract: AccountId::from_str(contract)?,
+            msg: msg_bytes,
+            funds,
+        }
+        .to_any()?;
+
+        let raw_tx = signing_client
+            .create_tx(wasm_tx, &self.chain_denom(), 500_000, 500_000u64, None)
```
keyleu, line 69: should fee amount and gas limit be hardcoded? I think we should run simulate_tx and then adjust these values accordingly. Example: https://github.com/Rhaki/cosmos-grpc-client/blob/19dcd72bc29213666000d5b25253284f68a382d9/cosmos-grpc-client/src/wallet.rs#L146

packages/chain-client-utils/Cargo.toml
bekauz: great point! done

packages/chain-client-utils/src/cosmos/wasm_client.rs
bekauz: also worth to mention, I think in the future we can make the client constructors even smaller by fetching more params (chain id, chain denom) from somewhere.

keyleu: I'm pretty sure wallets take this from the same place. Sometimes Keplr stops working from a chain because team modifies these values but forgets to update chain registry and therefore wrong values are taken, so this is indeed the right way of doing it

packages/chain-client-utils/src/cosmos/signing_client.rs
```diff
@@ -0,0 +1,104 @@
...
+
+        let tx_raw = sign_doc.sign(&self.signing_key)?;
+
+        let broadcast_tx_request = BroadcastTxRequest {
+            tx_bytes: tx_raw.to_bytes()?,
+            mode: BroadcastMode::Sync.into(),
+        };
+
+        // TODO: in the future we can consider auto-incrementing the account sequence number here.
+        // for now txs are signed with the account sequence number that gets fetched just in time.
```
keyleu, line 100: small note: I think current way is the right way (querying the number)

bekauz: cool, I'll keep the note in for now in case something starts going wrong with firing multiple execution requests with `tokio::join!`

PR #259, opened 2025-03-05
feat: add zk verification to authorization state

vlopes11:
This commit introduces a verifying key as optional attribute of the authorization layer.

Every deployed program will be bound to an authorization contract, and will optionally be a ZK program.

This commit introduces the skeleton for a ZK verification for such case. It adds a proof binary to a Cosm execute message (the other variants are currently unimplemented for ZK), and binds the authorization state to a ZK key for the shielded program.

It uses ark-groth16, an industry standard that will support the output proof of most of the zkVM implementations. As tech debt, we might want a separated library that will decouple such decisions from the concrete implementation, leaving all of the involved implementations (zkVM, verifier, msg encoding) decoupled on this crate.

## Summary by CodeRabbit

- **New Features**
	- Enhanced security with zero-knowledge proof verification integrated into authorization and messaging.
	- Introduced support for new cryptographic verification keys, improving error detection for proof failures.
	- Expanded functionality to process cryptographic messages with an updated message type.
	- Upgraded dependency libraries to bolster cryptographic operations and overall data integrity.
	- Added new properties for zero-knowledge verification keys in various schemas and data structures.

- **Bug Fixes**
	- Improved validation logic for message processing to ensure proof verification is correctly handled.

### DIFF
--- a/packages/authorization-utils/src/authorization.rs
+++ b/packages/authorization-utils/src/authorization.rs
@@ -65,7 +65,7 @@ impl AuthorizationInfo {
             subroutine: self.subroutine,
             priority: self.priority.unwrap_or_default(),
             state: AuthorizationState::Enabled,
-            zk_vk: None,
+            zk_vk: self.zk_vk,
         }
     }
 }


keyleu:
we have a script in /devtools to regenerate the schemas when APIs are changed

bekauz (approved):
looks good to me!

packages/authorization-utils/src/msg.rs
```diff
@@ -265,6 +266,7 @@ pub enum InternalAuthorizationMsg {
 #[cw_serde]
 pub enum ProcessorMessage {
     CosmwasmExecuteMsg { msg: Binary },
+    CosmwasmExecuteZKMsg { msg: Binary, proof: Binary },
```
keyleu, line 269: Can we rename it to CosmwasmExecuteZkMsg everywhere? or use serde rename on top of them? Mainly because this gets serialized/deserialized as cosmwasm_execute_z_k_msg which is kinda ugly. Would also require regenerating the schemas

keyleu (changes requested):

keyleu (approved):

keyleu (changes requested):
should probably close this PR

PR #271, opened 2025-03-18
feat: ICA IBC transfer library

keyleu:
Adds the ICA IBC transfer library which is the last library needed for the Vaults flow. The way it works is: the USDC will be sent via CCTP to an account on Noble (the remote account created by the Valence ICA). Then this library will be used to transfer these funds to other IBC connected chains.

Adds the corresponding unit/e2e tests and a doc section for it.

## Summary by CodeRabbit

- **New Features**
  - Introduced an interchain token transfer capability using the IBC protocol to enable remote transfers via interchain accounts.
- **Documentation**
  - Added comprehensive guides and diagrams detailing configuration and usage of the new transfer functionality.
- **Tests**
  - Implemented extensive testing and an end-to-end example to ensure reliable transfer operations and proper configuration management.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8598,6 +8598,7 @@ dependencies = [
  "valence-generic-ibc-transfer-library",
  "valence-ibc-utils",
  "valence-ica-cctp-transfer",
+ "valence-ica-ibc-transfer",
  "valence-interchain-account",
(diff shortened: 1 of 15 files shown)


bekauz (approved):
lgtm

PR #273, opened 2025-03-18
fix: lib account changes

Art3miX:
Tidy up the `LibraryAccountType` in our code to be better.

Fixed the next problem:

**Problem:** 
Library expected a String in one of its field, but we want to use an account (LibraryAccountType) there.
Previously it was not possible because the manager would take the id and replace it with the address, but the type would stay `LibraryAccountType`

The PR fixes this by providing another function: `to_raw_placeholder()`, which would provide a string placeholder, this placeholder will output:
1. In case the `LibraryAccountType` is already an `Addr`, this will output the address as a simple string.
2. In case the `LibraryAccountType` is an `AccountId` it will provide a placeholder string for the manager to replace and id with the instantiated account address, example: `|lib_acc_placeholder|:1` will be replaced to the address of account id 1.

Note: When using this, the account must be added to the link of this library as output account.

## Summary by CodeRabbit

- **Documentation**
  - Revised descriptions across configuration schemas to clarify how accounts and libraries are associated with identifiers, detailing the transition from temporary IDs to instantiated addresses.
- **Refactor**
  - Streamlined module organization and updated import paths to improve consistency and maintainability.
- **New Features**
  - Introduced enhanced placeholder handling for account configurations to support dynamic processing based on program instantiation.


packages/library-utils/src/library_account_type.rs
```diff
@@ -0,0 +1,273 @@
+use std::str::FromStr;
+
+use cosmwasm_schema::cw_serde;
+use cosmwasm_std::{Addr, StdError, StdResult};
+
+use crate::Id;
+
+pub const LIBRARY_ACCOUNT_RAW_PLACEHOLDER: &str = "|lib_acc_placeholder|";
+
+/// An helper type that is used to associate an account or library with an id
```
keyleu, line 10: nit: a helper

keyleu (dismissed):

packages/library-utils/src/library_account_type.rs
```diff
@@ -0,0 +1,273 @@
...
+    /// id replacement functionality of the manager.
+    /// Using this function will use a placeholder that can be replaced by the manager
+    /// to the instantiated address
+    pub fn to_raw_placeholder(&self) -> String {
+        let value = match self {
+            // If its an address, we can use it directly
+            LibraryAccountType::Addr(addr) => return addr.to_string(),
+            LibraryAccountType::AccountId(id) => id.to_string(),
+            LibraryAccountType::LibraryId(_) => {
+                panic!("Only accounts can use raw_placeholder functionality")
```
bekauz, line 55: maybe worth to return `StdResult` here as well to avoid `panic` &  be consistent with `to_string() / to_addr()`?

packages/library-utils/src/library_account_type.rs
```diff
@@ -0,0 +1,273 @@
...
+            LibraryAccountType::Addr(
+                input
+                    .trim_start_matches("{\"|library_account_addr|\":\"")
+                    .trim_end_matches("\"}")
+                    .parse()
+                    .expect("Failed parsing addr into LibraryAccountType"),
+            )
+        } else {
+            LibraryAccountType::Addr(input.to_string())
+        }
```
bekauz, line 97: some comment here would be helpful about why in case the `if input.starts_with("{\"|library_account_addr|\":\"")` check fails we still default to `LibraryAccountType::Addr`?

packages/library-utils/src/library_account_type.rs
```diff
@@ -0,0 +1,273 @@
...
+use crate::Id;
+
+pub const LIBRARY_ACCOUNT_RAW_PLACEHOLDER: &str = "|lib_acc_placeholder|";
+
+/// An helper type that is used to associate an account or library with an id
+/// When a program is not instantiated yet, ids will be used to reference accounts and libraries
+/// When a program is instantiated, the ids will be replaced by the instantiated addresses
+#[cw_serde]
+#[derive(Eq, PartialOrd, Ord)]
+pub enum LibraryAccountType {
```
bekauz, line 15: not sure if this is the right place for this fb but `LibraryAccountType` naming doesn't feel right.

to me `LibraryAccount` implies that we're dealing with accounts (that belong to some library), but there are variants like `LibraryId` which feels contradicting. perhaps can think of something more descriptive?

packages/library-utils/src/library_account_type.rs
Art3miX: The reason `to_string() / to_addr()` is using `StdResult` is because both are being used in contract environment.

`to_raw_placeholder()` should never be used in a contract only while building the program, there is no way to recover from this error.

I know it looks weird but as far as I know, this is actually following best practice.

packages/library-utils/src/library_account_type.rs
Art3miX: Any suggestions? @bekauz @keyleu @uditvira ?
I also don't like the name, but just couldn't think of something better that is not extremely long.
This type is a placeholder in library configs for accounts, so accounts are represented by ids while building the program, and when we instantiate it, those ids are getting replaced with their representing addresses.

keyleu:
need to regen schemas

packages/library-utils/src/library_account_type.rs
keyleu: since this is unfortunately coupled with all our CW libraries and is a breaking change for previous version APIs I wouldn't touch this unless we decide to somehow do a much deeper refactor.

keyleu (approved):

PR #275, opened 2025-03-19
fix: Vault completing withdraw requests when no balance on withdraw account

keyleu:
When there's no balance in the withdraw account and the withdraw request is completed by a solver using the batch completion, if there's no balance in the withdraw account the withdraw request is still being removed from the array.

Fixed that so that this cannot happen, by restoring the request if the transfer failed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved withdrawal error handling so that if an asset transfer fails, your withdrawal request is preserved for retry.

- **Tests**
  - Added a new test scenario to verify batch withdrawal behavior when available funds are depleted.

- **New Features**
  - Introduced a token burning capability in the `MockERC20` contract to allow token holders to reduce their total supply.

### DIFF
--- a/solidity/src/libraries/ValenceVault.sol
+++ b/solidity/src/libraries/ValenceVault.sol
@@ -775,20 +775,17 @@ contract ValenceVault is
             assetsToWithdraw = shares.mulDiv(updateInfo.withdrawRate, ONE_SHARE, Math.Rounding.Floor);
         }
 
-        // Delete request before transfer to prevent reentrancy
-        delete userWithdrawRequest[owner];
-
         // Prepare the transfer
         bytes memory transferCalldata = abi.encodeCall(IERC20.transfer, (request.receiver, assetsToWithdraw));
 
         // Execute transfer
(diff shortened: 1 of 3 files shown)


bekauz (approved):
this looks great! i believe it closes #276 , right?

solidity/src/libraries/ValenceVault.sol
```diff
@@ -787,6 +787,8 @@ contract ValenceVault is
             return WithdrawResult(true, assetsToWithdraw, request.solverFee, "");
         } catch {
             if (revertOnFailure) revert("Asset transfer failed");
+            // Restore request if transfer fails
+            userWithdrawRequest[owner] = request;
```
Art3miX, line 791: Seems to be unneeded write.

Can't we just `delete userWithdrawRequest[owner];` if the transfer was correct? 
This should be safe because `completeWithdraws` already have the `nonReentrant` guard

keyleu: you're right, both have the guard so we can just delete it after transfer succeeds.
was confused by the comment on the delete

PR #276, opened 2025-03-19
fix: dont complete withdraws when there was no update performed

### LINKED ISSUE #265: bug: vault redeem request processing without update
# description

user attempting to finalize a withdrawal without a vault update between the redeem and withdraw completion requests results in loss of funds.

## steps to reproduce

1. user deposits funds into the vault, gets vault shares in return
2. user initiates the redeem request which causes the vault to burn the user shares being redeemed
3. user triggers the withdraw completion request, causing the vault to finalize the user withdraw request but not crediting the user with underlying funds from the withdraw account

keyleu:
closes #265 

When a user was trying to complete a withdraw request and there was no update performed before that, when fetching from the updateInfos array it was using default values (0) because the update didn't exist.

This caused all the calculations to be 0 and the withdraw to succeed sending 0 tokens to the user and his withdraw request removed.

This PR doesn't allow this from happening. If there's no update, the withdraw cannot be completed and reverts/continues to the next one.

## Summary by CodeRabbit

- **New Features**
  - Enhanced the withdrawal process to clearly indicate when an update is missing, ensuring users get appropriate feedback.
- **Tests**
  - Added tests to confirm that withdrawal attempts without the necessary updates are correctly blocked.

### DIFF
--- a/solidity/src/libraries/ValenceVault.sol
+++ b/solidity/src/libraries/ValenceVault.sol
@@ -40,6 +40,7 @@ contract ValenceVault is
     error WithdrawRequestNotFound();
     error SolverNotAllowed();
     error WithdrawNotClaimable();
+    error NoUpdateForRequest();
     error SolverFeeTransferFailed();
     error FeeExceedsUint128();
     error OnlyOwnerCanUnpause();
@@ -738,6 +739,12 @@ contract ValenceVault is
         uint256 _redemptionRate = redemptionRate;
         UpdateInfo memory updateInfo = updateInfos[request.updateId];
         uint256 assetsToWithdraw;
+
+        // Check if there was an update for this request
+        if (updateInfo.timestamp == 0) {
+            if (revertOnFailure) revert NoUpdateForRequest();
+            return WithdrawResult(false, 0, 0, "Update not found");
+        }
 
         // The current withdrawRate is the redemption rate minus the update withdraw fee
         uint256 currentWithdrawRate = _redemptionRate.mulDiv(BASIS_POINTS - updateInfo.withdrawFee, BASIS_POINTS);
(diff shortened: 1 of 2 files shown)


solidity/src/libraries/ValenceVault.sol
```diff
@@ -739,6 +740,12 @@ contract ValenceVault is
         UpdateInfo memory updateInfo = updateInfos[request.updateId];
         uint256 assetsToWithdraw;
 
+        // Check if there was an update for this request
+        if (updateInfo.timestamp == 0) {
```
bekauz, line 744: can this timestamp be outdated (leftover) from a previous epoch?

keyleu: Nope it's an absolute timestamp, so if it's 0 we know that update didn't happen

Art3miX:
I think we might have a bigger issue here, each request has a `claimTime`, if the update didn't happen at the scheduled time, and got a delay, we might have a delay here.

A solution might be to move the `claimTime` to the update info instead of each request, so no matter when update happened the claimTime will be correct.

keyleu:
> I think we might have a bigger issue here, each request has a `claimTime`, if the update didn't happen at the scheduled time, and got a delay, we might have a delay here.
> 
> A solution might be to move the `claimTime` to the update info instead of each request, so no matter when update happened the claimTime will be correct.

There are multiple options that we have here.
We can use the rate of the previous update if the update didn't happen.
We can move the claimTime to the update (but how would it work then? all withdraws for the previous update are claimable at the same time? what if the update doesn't happen for a while?)
Or we can just apply the current fix where people can't claim unless there is an update.

In any case I suggest applying this patch for now and have a discussion and if we decide to modify the behavior doing it in a separate PR.

bekauz:
agree on getting this merged now as it does solve #265, delays can be investigated next. I also felt like something could go weird there; on a high level it feels like update status should be tracked per epoch and not per individual user. would make any delay resolution easier as there could only be one.

PR #286, opened 2025-03-26
feat: StargateTransfer EVM library

keyleu:
Adds the StargateTransfer library which leverages Stargate V2 (build on top of LayerZero v2) - more info on https://stargateprotocol.gitbook.io/stargate/v2-developer-docs - to send tokens across domains.

Supports transferring both ERC20 and Native tokens and sending fixed/full amounts from the input account.
Also supports both Bus and Taxi mode, including tests for both against mainnet contracts.
Includes both unit tests with Stargate mocking and tests against the mainnet contracts to veriffy that interaction is correct.

Note: fees are always paid in Native token.

Also adds the encoding part to the EVM encoder.

The doc section is also included for @uditvira to review.

## Summary by CodeRabbit

- **New Features**
	- Introduced a cross-chain transfer option using the Stargate protocol, enabling seamless asset transfers across networks.
	- Added support for the new `StargateTransfer` functionality, including configuration for transfers and error handling.
	- Deployed updated deployment scripts and smart contracts for executing both native and ERC20 token transfers.
- **Documentation**
	- Expanded and updated guides and flow diagrams to include the new Stargate Transfer feature for clearer user understanding.
	- Added new documentation entries for the Stargate Transfer library, detailing its configuration and usage.


contracts/encoders/evm-encoder/src/libraries/stargate_transfer.rs
```diff
@@ -0,0 +1,106 @@
...
+    /// The recipient for the library. Bytes32 representation of the address in solidity.
+    pub recipient: Binary,
+    /// The destination domain to transfer to
+    pub destination_domain: u32,
+    /// Address of Stargate Pool contract.
+    pub stargate_address: String,
+    /// The address of the token to transfer.
+    pub transfer_token: String,
+    /// Amount to transfer. Setting this to 0 will transfer the entire balance.
+    pub amount: Uint256,
```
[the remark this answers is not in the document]
keyleu: Yes, it's a low impact change, but then we wouldnt be able to updated for bigger amounts than that.

keyleu:
> Looks good; would be nice if we could have some tests on the rust end too

I have an issue open to add e2e tests for all config updates. It's low priority at the moment because we might not/never use this encoder with hyperlane, also we might not do updateConfigs through rust code.

There is an e2e test that indeed has tests executing the libraries from rust in ethereum_integration_tests so everything except updateConfigs is tested.

contracts/encoders/evm-encoder/src/libraries/stargate_transfer.rs
```diff
@@ -0,0 +1,106 @@
...
+use super::{get_update_ownership_call, get_update_processor_call};
+
+// We need to define a config and functions for this library as we don't have a CosmWasm equivalent
+#[cw_serde]
+/// Struct representing the library configuration.
+pub struct LibraryConfig {
+    /// The input address for the library.
+    pub input_addr: LibraryAccountType,
+    /// The recipient for the library. Bytes32 representation of the address in solidity.
+    pub recipient: Binary,
```
bekauz, line 20: wyt if we make a `newtype` wrapper for evm addresses in cw context? something like `struct EvmAddress(pub Binary)`.

seems like doing operations like conversions into fixed sizes is becoming a pattern so it could be implemented on that newtype level to be reused across other libs as well

```rust
 let recipient_fixed: [u8; 32] = config.recipient.as_slice().try_into().map_err(|e| {
        StdError::generic_err(format!(
            "Error converting mint recipient to fixed size: {}",
            e
        ))
    })?;

```

docs/src/libraries/evm/stargate_transfer.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Functions
+
+| Function     | Parameters | Description                                                                                                   |
+| ------------ | ---------- | ------------------------------------------------------------------------------------------------------------- |
+| **Transfer** | -          | Transfer funds from the configured **input account** to the **recipient** on the **destination domain**. |
+
+## Configuration
+
+The library is configured on deployment using the `StargateTransferConfig` type. A list of the supported Stargate destination domains that can be used in the `destinationDomain` field can be found [here](https://stargateprotocol.gitbook.io/stargate/v2-developer-docs/technical-reference/mainnet-contracts).
```
bekauz, line 45:
```suggestion
The library is configured on deployment using the `StargateTransferConfig` type. A list of supported Stargate destination domains that can be used in the `destinationDomain` field can be found [here](https://stargateprotocol.gitbook.io/stargate/v2-developer-docs/technical-reference/mainnet-contracts).
```

solidity/script/StargateTransfer.script.sol
```diff
@@ -0,0 +1,168 @@
...
+        // Deploy a new BaseAccount contract
+        inputAccount = new BaseAccount(owner, new address[](0));
+
+        // Mint some USDC tokens to the BaseAccount
+        MockERC20 usdc = MockERC20(USDC_ADDR);
+        usdc.mint(address(inputAccount), tokenAmount); // Mint the USDC amount we want to transfer
+
+        // Send some ETH to the BaseAccount
+        // We are going to transfer first a fixed amount and then the remaining full amount to test both scenarios
+        vm.deal(address(inputAccount), ethAmount * 2);
```
bekauz, line 59: not something to be addressed but I'm curious why foundry chose to name their env methods with names like `deal`, `prank`, and `etch`. on a first glance feels like unnecessary learning curve? is it just to show that these calls aren't part of the mainnet evm flow?

solidity/src/libraries/StargateTransfer.sol
```diff
@@ -0,0 +1,257 @@
...
+
+        // Ensure the starport address is valid (non-zero).
+        if (decodedConfig.stargateAddress == IStargate(address(0))) {
+            revert("Stargate address can't be zero address");
+        }
+
+        // Ensure the transfer token and the token address in the stargate address match.
+        if (decodedConfig.stargateAddress.token() != decodedConfig.transferToken) {
+            revert("Token address does not match the stargate token address");
+        }
```
bekauz, line 80: if we can derive the `transferToken` from `config.stargateAddress`, maybe we can exclude the `transferToken` from `StargateTransferConfig`?

bekauz (commented):
lgtm, just a few points

contracts/encoders/evm-encoder/src/libraries/stargate_transfer.rs
keyleu: Will do! Will prob call it Bytes32Address or something like that because it's not only for EVM

solidity/src/libraries/StargateTransfer.sol
keyleu: 2 reasons imo.
1. All our ibc transfer/bridge libraries have the denom/token on the config. So for consistency and to avoid confusion on "why not here?"
2. For confirmation. Not inferring it will make sure that the creator knows what he's transferring and didn't make a mistake that can be painful to debug later

solidity/script/StargateTransfer.script.sol
keyleu: If I had to guess it would be to keep them short and concise to have cleaner tests, instead of more verbose like impersonate or swapCode. Feels more like hackery vibes. But really not sure, just guessing

contracts/encoders/evm-encoder/src/libraries/stargate_transfer.rs
keyleu: Added a new Bytes32Address wrapped around Binary to re-use the operation.
Included also a constructor from binary that does the check that indeed you are passing 32  bytes when creating it so that we can't create invalid ones

uditvira:
Docs lgtm!

uditvira (approved):

PR #292, opened 2025-03-28
add skip swap library with off-chain strategist component

hxrts:
this is a library allows setting a desired route that, when funds are deposited, an authorized off chain strategist can query the skip api, return a message, which is validated against the route.

making this a draft PR because I need to get lunch and it requires another pass to make sure it's fully there. all the unit tests and the mock integration test work though.


Art3miX:
The contract is missing our library specifics, like `LibraryConfig` for example, and helper execute message.

PR #296, opened 2025-04-01
feat: add pfm to ica ibc transfer library

keyleu:
adds the PFM functionality to the ICA IBC transfer library in the same way it works for the IBC generic transfer and IBC neutron transfer libraries. This also adds the memo field in the config in case we want to send the IBC transfer with a memo for transfers without PFM. Updated docs with new fields and tests to confirm it works as expected.

## Summary by CodeRabbit

- **New Features**
  - Introduced optional memo field and flexible multi-hop routing in interchain transfers.
  - Expanded blockchain network support with new configurations for multiple chains, including "gaia", "neutron", "noble", and "osmosis".

- **Documentation**
  - Updated user guides to include new fields and configurations for the Packet Forwarding Middleware and multi-hop transfers.


docs/src/libraries/cosmwasm/ica_ibc_transfer.md
```diff
@@ -55,4 +59,23 @@ pub struct RemoteChainInfo {
...
     pub ibc_transfer_timeout: Option<u64>,
 }
+
+// Configuration for a multi-hop transfer using the Packet Forwarding Middleware
+struct PacketForwardMiddlewareConfig {
+  // Channel ID from the source chain to the intermediate chain
+  local_to_hop_chain_channel_id: String,
+  // Channel ID from the intermediate to the destination chain
+  hop_to_destination_chain_channel_id: String,
+  // Temporary receiver address on the intermediate chain
```
uditvira, line 69:
```suggestion
  // Temporary receiver address on the intermediate chain. Typically this is set to an invalid address so the entire transaction will revert if the transfer fails
```

contracts/libraries/ica-ibc-transfer/README.md
```diff
@@ -49,4 +53,14 @@ pub struct RemoteChainInfo {
...
 }
+
+// Configuration for a multi-hop transfer using the Packet Forwarding Middleware
+struct PacketForwardMiddlewareConfig {
+  // Channel ID from the source chain to the intermediate chain
+  local_to_hop_chain_channel_id: String,
+  // Channel ID from the intermediate to the destination chain
+  hop_to_destination_chain_channel_id: String,
+  // Temporary receiver address on the intermediate chain. Typically this is set to an invalid address so the entire transaction will revert if the forwarding fails
+  hop_chain_receiver_address: String,
```
bekauz, line 64: wyt if we made this `Option<String>` and defaulted to `"pfm"` string if it wasn't passed? I think that's what the pfm docs suggest passing to the hop receiver address anyways

contracts/libraries/ica-ibc-transfer/src/contract.rs
```diff
@@ -67,25 +66,58 @@ mod functions {
...
+                            .nanos(),
+                        memo: to_json_string(&PacketMetadata {
+                            forward: Some(ForwardMetadata {
+                                receiver: cfg.receiver.clone(),
+                                port: "transfer".to_string(),
+                                // hop chain to final receiver chain channel
+                                channel: pfm_config.hop_to_destination_chain_channel_id.to_string(),
+                            }),
+                        })?,
+                    },
```
bekauz, line 120: can we not reuse the `ibc_send_message()` helper from  `packages/ibc-utils/src/generic.rs` here?

keyleu: nope, it's using the cosmwasm-std IbcMsg API there

contracts/libraries/ica-ibc-transfer/README.md
keyleu: let me do this in a separate PR as it affects the other IBC contracts (generic transfer and neutron transfer) too because PacketForwardMiddlewareConfig is used in all of them

keyleu: I'll create an issue for it

contracts/libraries/ica-ibc-transfer/src/contract.rs
keyleu: Also moving this also there is a bit annoying because we depend on different cosmos-sdk-proto versions. We need 0.20.0 for neutron proto types because we need Coin from there and here we need to use a later version to use with IBC to pass the memo, so I rather keep it here as it's probably the only place where we are eventually going to do something like this.

bekauz: fair enough

bekauz (approved):

PR #299, opened 2025-04-01
feat: mock cctp relayer

### LINKED ISSUE #279: feat: mock cctp relayer
# description

in order to simulate a full e2e flow involving cctp transfers between different domains we need a relayer mocking the circle's attestation service.

initially we should focus on supporting noble <-> ethereum flows.

## general approach

to keep things as close to mainnet, the relayer should be decoupled from the test flow and run independently in the background.

relayer should operate bidirectionally (1 instance handles noble -> evm and evm -> noble).

after picking up the burn event on the origin domain, a mint should be triggered on the destination domain. receiver address and any other relevant information should be inferred from the origin domain.

## future

in the future it would be useful to have the relayer be configurable with regards to delays, failures and other weird edge cases.

bekauz:
closes #279 

# Description

This pr introduces the initial implementation of relaying CCTP requests between Ethereum and Noble chains.

On Ethereum side, the process tracks the events emitted by the `MockTokenMessenger` contract, processes them, and uses the event information to mint the corresponding funds on Noble.

On Noble side, the process queries the block results searching for `circle.cctp.v1.DepositForBurn` event, processes it, and uses that information to mint the corresponding funds on Ethereum.

## Summary by CodeRabbit

- **New Features**
  - Introduced an advanced cross-chain relayer to streamline token transfers and event processing.
  - Added functionality to dynamically retrieve updated blockchain RPC endpoints.

- **Refactor**
  - Improved account management during transfers with refined logging and error handling.
  - Enhanced state tracking and integration for smoother cross-chain operations.
  - Updated the method of obtaining gRPC addresses and ports for better structure and error handling.

- **Bug Fixes**
  - Resolved issues with gRPC address retrieval and improved error handling in related functions.

- **Style**
  - Standardized configuration and package formatting for increased consistency.


bekauz:
@keyleu I think this is ready, seem to be working pretty consistently in the e2e run. Added a follow-up issue for shutdowns in #300 as it didn't feel urgent right now.

e2e/src/utils/mock_cctp_relayer.rs
```diff
@@ -0,0 +1,329 @@
...
+
+use super::{
+    parse::get_grpc_address_and_port_from_logs, ADMIN_MNEMONIC, DEFAULT_ANVIL_RPC_ENDPOINT,
+    NOBLE_CHAIN_DENOM,
+};
+
+pub struct MockCctpRelayer {
+    eth_client: EthereumClient,
+    noble_client: NobleClient,
+    state: Arc<Mutex<RelayerState>>,
```
keyleu, line 40: I'm a little bit confused of the need of an Arc Mutex here since the relayer is a single threaded application and operations on the state happen sequentially unless I'm missing something.
Couldn't it be simplified into something like this?:
```rust
pub struct MockCctpRelayer {
    eth_client: EthereumClient,
    noble_client: NobleClient,
    state: RelayerState,
}

// Change from:
async fn poll_noble(&self, ...) {
    let mut state = self.state.lock().await;
    // ...
    state.last_noble_block = current_block;
}

// To:
async fn poll_noble(&mut self, ...) 
    // ...
    self.state.last_noble_block = current_block;
}
```

Now this would prevent the main thread to have access to the relayer at all but I guess that's what we want right? We are just mocking some external infra that we wouldn't have access to anyways.

packages/chain-client-utils/Cargo.toml
```diff
@@ -1,33 +1,36 @@
 [package]
-name       = "valence-chain-client-utils"
```
keyleu, line 2: nit: run just_toml

packages/chain-client-utils/src/cosmos/base_client.rs
```diff
@@ -88,6 +94,20 @@ pub trait BaseClient: GrpcSigningClient {
...
+    async fn block_results(
+        &self,
+        rpc_addr: &str,
+        height: u32,
+    ) -> Result<cosmrs::rpc::endpoint::block_results::Response, StrategistError> {
+        let client = HttpClient::new(rpc_addr)?;
+
+        let height = Height::from(height);
+
+        let results = client.block_results(height).await.unwrap();
```
keyleu: should maybe not unwrap here and propagate error?

packages/chain-client-utils/src/cosmos/base_client.rs
```diff
@@ -88,6 +94,20 @@ pub trait BaseClient: GrpcSigningClient {
         Ok(block_header)
     }
 
+    async fn block_results(
+        &self,
+        rpc_addr: &str,
+        height: u32,
+    ) -> Result<cosmrs::rpc::endpoint::block_results::Response, StrategistError> {
```
keyleu, line 101: Should we have the concept of StrategistError in the client-utils package?

packages/chain-client-utils/src/cosmos/base_client.rs
bekauz: done

keyleu (approved):

e2e/src/utils/mock_cctp_relayer.rs
```diff
@@ -0,0 +1,349 @@
...
+            if self.state.eth_processed_events.insert(event_id) {
+                info!("[CCTP MOCK RELAY] picked up CCTP transfer event on Ethereum");
+
+                let alloy_log =
+                    Log::new(log.address(), log.topics().into(), log.data().clone().data)
+                        .unwrap_or_default();
+
+                let deposit_for_burn_log = DepositForBurn::decode_log(&alloy_log, false)?;
+                self.mint_noble(deposit_for_burn_log).await?;
+            }
```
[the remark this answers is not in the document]
bekauz: I don't think we can rely on them being ordered but in the future we can look into how the query filter can be made more narrow. I think for this case it shouldn't cause much issues as it's just iterating over a vector which should be fine.

PR #307, opened 2025-04-07
feat: AAVE PositionManager library

keyleu:
Adds the AAVEPositionManager library which leverages AAVE V3 for lending/borrowing/repaying/withdrawing.

Includes a doc section that explains all the flows, @uditvira.

This library works a bit differently, since it has multiple "intermediate operations". The input account is the holder of all intermediate tokens (aTokens, debt tokens...) and the output account is only used for the withdraw step, where the tokens withdrawn are sent back to it.

The PR includes the encoder section, unit tests and e2e tests with the AAVE protocol forking mainnet.

## Summary by CodeRabbit

- **New Features**
	- Introduced AAVE Position Manager functionality, enabling users to interact with AAVE v3 for supply, borrow, withdraw, and repayment operations.
- **Documentation**
	- Added new guides and flowcharts detailing the AAVE Position Manager usage and configuration.
- **Tests**
	- Implemented comprehensive testing, including simulations and mock scenarios, to ensure robust functionality.
- **Chores**
	- Updated dependency management and modular organization to support the AAVE integration.

### DIFF
--- a/contracts/encoders/evm-encoder/src/libraries/aave_position_manager.rs
+++ b/contracts/encoders/evm-encoder/src/libraries/aave_position_manager.rs
@@ -1,0 +1,119 @@
+use alloy_sol_types::{SolCall, SolValue};
+use cosmwasm_schema::cw_serde;
+use cosmwasm_std::{Binary, StdError, StdResult, Uint256};
+
+use valence_encoder_utils::libraries::{
+    aave_position_manager::solidity_types::{
+        borrowCall, repayCall, repayWithATokensCall, supplyCall, withdrawCall,
+    },
+    updateConfigCall,
+};
+use valence_library_utils::{msg::ExecuteMsg, LibraryAccountType};
+
+use crate::parse_address;
+
+use super::{get_update_ownership_call, get_update_processor_call};
+
+// We need to define a config and functions for this library as we don't have a CosmWasm equivalent
+#[cw_serde]
+/// Struct representing the library configuration.
+pub struct LibraryConfig {
+    /// The input address for the library.
+    pub input_addr: LibraryAccountType,
+    /// The output address for the library.
+    pub output_addr: LibraryAccountType,
+    /// The AAVE pool address.
+    pub aave_pool: String,
+    /// The supply asset token address.
+    pub supply_asset: String,
+    /// The borrow asset token address.
+    pub borrow_asset: String,
+    /// The referral code if applicable.
+    pub referral_code: Option<u16>,
+}
+
+#[cw_serde]
+/// Enum representing the different function messages that can be sent.
+pub enum FunctionMsgs {
+    /// Message to supply tokens.
+    Supply { amount: Uint256 },
+    /// Message to borrow tokens.
+    Borrow { amount: Uint256 },
+    /// Message to withdraw tokens.
+    Withdraw { amount: Uint256 },
+    /// Message to repay tokens.
+    Repay { amount: Uint256 },
+    /// Message to repay with aTokens.
+    RepayWithATokens { amount: Uint256 },
+}
+
+type StargateTransferMsg = ExecuteMsg<FunctionMsgs, LibraryConfig>;
+
+pub fn encode(msg: &Binary) -> StdResult<Vec<u8>> {
+    // Extract the message from the binary and verify that it parses into a valid json for the library
+    let msg: StargateTransferMsg = serde_json::from_slice(msg.as_slice()).map_err(|_| {
+        StdError::generic_err("Message sent is not a valid message for this library!".to_string())
+    })?;
+
+    match msg {
+        ExecuteMsg::ProcessFunction(function) => match function {
+            FunctionMsgs::Supply { amount } => {
+                let supply_call = supplyCall {
+                    amount: alloy_primitives::U256::from_be_bytes(amount.to_be_bytes()),
+                };
+                Ok(supply_call.abi_encode())
+            }
+            FunctionMsgs::Borrow { amount } => {
+                let borrow_call = borrowCall {
+                    amount: alloy_primitives::U256::from_be_bytes(amount.to_be_bytes()),
+                };
+                Ok(borrow_call.abi_encode())
+            }
+            FunctionMsgs::Withdraw { amount } => {
+                let withdraw_call = withdrawCall {
+                    amount: alloy_primitives::U256::from_be_bytes(amount.to_be_bytes()),
+                };
+                Ok(withdraw_call.abi_encode())
+            }
+            FunctionMsgs::Repay { amount } => {
+                let withdraw_call = repayCall {
+                    amount: alloy_primitives::U256::from_be_bytes(amount.to_be_bytes()),
+                };
+                Ok(withdraw_call.abi_encode())
+            }
+            FunctionMsgs::RepayWithATokens { amount } => {
+                let withdraw_call = repayWithATokensCall {
+                    amount: alloy_primitives::U256::from_be_bytes(amount.to_be_bytes()),
+                };
+                Ok(withdraw_call.abi_encode())
+            }
+        },
+        ExecuteMsg::UpdateConfig { new_config } => {
+            // Parse addresses
+            let input_account = parse_address(&new_config.input_addr.to_string()?)?;
+            let output_account = parse_address(&new_config.output_addr.to_string()?)?;
+            let aave_pool_address = parse_address(&new_config.aave_pool)?;
+            let supply_asset = parse_address(&new_config.supply_asset)?;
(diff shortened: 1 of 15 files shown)


solidity/test/libraries/AavePositionManager.t.sol
```diff
@@ -0,0 +1,356 @@
...
+        // Mint tokens to input account
+        uint256 balance = 100 * 10 ** 18;
+        vm.prank(owner);
+        supplyToken.mint(address(inputAccount), balance);
+
+        // Execute supply with amount larger than balance (should fail)
+        vm.prank(processor);
+        vm.expectRevert("Insufficient supply asset balance");
+        aavePositionManager.supply(200 * 10 ** 18);
+    }
```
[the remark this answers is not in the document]
keyleu: done in e2e tests, the mock contract doesn't do anything

solidity/test/libraries/AavePositionManager.t.sol
```diff
@@ -0,0 +1,356 @@
...
+
+    // ============== Repay With ATokens Tests ==============
+
+    function testRepayWithATokens() public {
+        // Execute repayWithATokens as processor
+        uint256 amount = 100 * 10 ** 18;
+        vm.prank(processor);
+        aavePositionManager.repayWithATokens(amount);
+    }
```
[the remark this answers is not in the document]
keyleu: done in e2e tests

contracts/encoders/evm-encoder/src/libraries/aave_position_manager.rs
```diff
@@ -0,0 +1,119 @@
...
+// We need to define a config and functions for this library as we don't have a CosmWasm equivalent
+#[cw_serde]
+/// Struct representing the library configuration.
+pub struct LibraryConfig {
+    /// The input address for the library.
+    pub input_addr: LibraryAccountType,
+    /// The output address for the library.
+    pub output_addr: LibraryAccountType,
+    /// The AAVE pool address.
+    pub aave_pool: String,
```
bekauz, line 26: feels like it's clear that it's an aave pool given that its coming from `aave_position_manager.rs`
```suggestion
    pub pool: String,
```

solidity/src/libraries/AavePositionManager.sol
```diff
@@ -0,0 +1,241 @@
...
+     * @param inputAccount The account from which transactions will be initiated
+     * @param outputAccount The account that will receive withdrawals. Can be the same as inputAccount.
+     * @param supplyAsset Address of the token to supply to Aave
+     * @param borrowAsset Address of the token to borrow from Aave
+     * @param referralCode Referral code for Aave protocol (if applicable - 0 if the action is executed directly by the user, without any middle-men)
+     */
+    struct AavePositionManagerConfig {
+        IPool aavePoolAddress;
+        Account inputAccount;
+        Account outputAccount;
```
Art3miX, line 29: Would suggest against using `Account` and go for a specific type of account (`BaseAccount`), mainly because choosing what account type should be used is based on functionality the library needs from the account, `BaseAccunt` is the default unless the library needs something else, so there should not be 2 possible options here.

contracts/encoders/evm-encoder/src/libraries/aave_position_manager.rs
```diff
@@ -0,0 +1,119 @@
...
+    /// Message to supply tokens.
+    Supply { amount: Uint256 },
+    /// Message to borrow tokens.
+    Borrow { amount: Uint256 },
+    /// Message to withdraw tokens.
+    Withdraw { amount: Uint256 },
+    /// Message to repay tokens.
+    Repay { amount: Uint256 },
+    /// Message to repay with aTokens.
+    RepayWithATokens { amount: Uint256 },
```
bekauz, line 47: do we need to highlight the `aTokens` part here? if i understand correctly it's a tokenized claim on the shares, maybe we can name this something like `RepayWithShares`?

solidity/src/libraries/AavePositionManager.sol
```diff
@@ -0,0 +1,241 @@
...
+        // Get the current balance of the supply asset in the input account
+        uint256 balance = IERC20(storedConfig.supplyAsset).balanceOf(address(storedConfig.inputAccount));
+
+        // Check if balance is zero
+        if (balance == 0) {
+            revert("No supply asset balance available");
+        }
+
+        // If amount is 0, use the entire balance
+        uint256 amountToSupply = amount == 0 ? balance : amount;
```
Art3miX, line 105: I think its confusing when some functions accept `0` as `all` and others accept `Type.max` as all, we should choose 1 consistent method.

contracts/encoders/evm-encoder/src/libraries/aave_position_manager.rs
bekauz, line 1: maybe we can add some comments pointing to the relevant aave codebase/docs in this file? would make life easier in the future

solidity/script/AavePositionManager.script.sol
```diff
@@ -0,0 +1,303 @@
...
+    address constant AUSDC_ADDR = 0x98C23E9d8f34FEFb1B7BD6a91B7FF122F4e16F5c;
+    address constant DAI_ADDR = 0x6B175474E89094C44Da98b954EedeAC495271d0F;
+    address constant ADAI_ADDR = 0x018008bfb33d285247A21d44E50697654f754e63;
+    address constant VARIABLE_DEBT_DAI = 0xcF8d0c70c850859266f5C338b38F9D663181C314;
+
+    // Let's use a whale to transfer some tokens as we need to fund the account
+    // Tried replacing the runtime code of the tokens but it messed up with AAVE's calculations because tokens had different decimals
+    // Binance hot wallet - huge USDC and DAI holdings
+    address constant USDC_WHALE = 0x28C6c06298d514Db089934071355E5743bf21d60;
+    address constant DAI_WHALE = 0x28C6c06298d514Db089934071355E5743bf21d60;
```
bekauz, line 25: lol, nice

solidity/src/libraries/AavePositionManager.sol
```diff
@@ -0,0 +1,241 @@
...
+    /**
+     * @dev Updates the AavePositionManager configuration.
+     * Only the contract owner is authorized to call this function.
+     * @param _config New encoded configuration parameters.
+     */
+    function updateConfig(bytes memory _config) public override onlyOwner {
+        // Validate and update the configuration.
+        config = validateConfig(_config);
+    }
+}
```
Art3miX, line 241: Should we also add fallback?

```solidity
/// @notice Fallback function that reverts all calls to non-existent function
/// @dev Called when no other function matches the function signature
/// @dev Add this to your contract to explicitly fail calls to wrong/non-existent 
fallback() external {
    revert("Function not found");
}
```

solidity/src/libraries/AavePositionManager.sol
```diff
@@ -0,0 +1,241 @@
...
+            revert("No borrow asset balance available");
+        }
+
+        // If amount is 0, use the entire balance
+        uint256 amountToRepay = amount == 0 ? balance : amount;
+
+        // Check if there's enough balance for the requested amount
+        if (balance < amountToRepay) {
+            revert("Insufficient borrow asset balance");
+        }
```
bekauz, line 195: clean

Art3miX (commented):
few nits

bekauz (commented):
this looks clean, lgtm!

only feedback I have is about aave-specific terminology being used in a few places. maybe we can stick to generic terms where possible so that other lending protocol integrations would be easier and eventually lead us to having common traits?

solidity/src/libraries/AavePositionManager.sol
keyleu: changed

solidity/src/libraries/AavePositionManager.sol
keyleu: changed to 0 for the two that use max

solidity/src/libraries/AavePositionManager.sol
keyleu: its inherited from Library

contracts/encoders/evm-encoder/src/libraries/aave_position_manager.rs
keyleu: changed it!

contracts/encoders/evm-encoder/src/libraries/aave_position_manager.rs
keyleu: all links to AAVE are in the doc section, keeping them always there so it's the place to go

bekauz (approved):

Art3miX (approved):

PR #309, opened 2025-04-08
fix: increase time between local-ic query and logs check

keyleu:
Sometimes it takes longer for logs to get created once we query local-ic, specially when we have 3+ chains. To avoid killing it even though it was successfully started, increase the time before we check that logs are there

## Summary by CodeRabbit

- **Chores**
  - Adjusted the local service startup routine by extending the wait period between retry attempts, enhancing stability during initialization.


Pacman99 (approved):
Sounds good

PR #312, opened 2025-04-10
feat: BalancerV2Swap library

keyleu:
Adds the BalancerV2Swap library which leverages Balancer V2 (https://docs-v2.balancer.fi/) to perform swaps and multi-hop swaps.

Includes: library with unit tests for validations and e2e tests for protocol interaction (e2e tests performed on BEX which is a balancer fork on Berachain to confirm library is compatible), encoding for interaction from CW, and a slightly more extended doc section to describe all parameters for the swapping (separated by calls).

This is the first library that has multiple parameters in its function calls, to make it more flexible for swapping and also because storing all the information in the config was prohibitively expensive compared to other libraries (for multi-hop swaps we have multiple arrays for example). It's also our first swap library so let's take a deeper look at the API used and the naming.

## Summary by CodeRabbit

- **New Features**
  - Introduced Balancer V2 Swap functionality, enabling secure single and multi-hop token swap operations.
  
- **Documentation**
  - Added new documentation detailing the Balancer V2 Swap functionality, including usage instructions and flow diagrams.
  - Updated the documentation summary to include Balancer V2 Swap.

- **Tests**
  - Added comprehensive tests and mock contracts to simulate and verify Balancer V2 Swap operations, ensuring robust performance.

### DIFF
--- a/contracts/encoders/evm-encoder/src/libraries/balancer_v2_swap.rs
+++ b/contracts/encoders/evm-encoder/src/libraries/balancer_v2_swap.rs
@@ -35,7 +35,7 @@ pub enum FunctionMsgs {
         token_in: String,
         /// The address of the token to swap to
         token_out: String,
-        /// The amount to swap from the first token
+        /// The amount to swap from the first token. If set to 0, the library will use the entire balance of the input account
         amount_in: Uint256,
         /// The minimum amount to receive from the last token
         min_amount_out: Uint256,
@@ -53,7 +53,7 @@ pub enum FunctionMsgs {
         /// List of tokens to swap through
         /// The first token is the one to swap from, the last one is the one to swap to
         tokens: Vec<String>,
-        /// The amount to swap from the first token
+        /// The amount to swap from the first token. If set to 0, the library will use the entire balance of the input account
         amount_in: Uint256,
         /// The minimum amount to receive from the last token
         min_amount_out: Uint256,
--- a/contracts/encoders/evm-encoder/src/libraries/balancer_v2_swap.rs
+++ b/contracts/encoders/evm-encoder/src/libraries/balancer_v2_swap.rs
@@ -1,0 +1,154 @@
+use alloy_sol_types::{SolCall, SolValue};
+use cosmwasm_schema::cw_serde;
+use cosmwasm_std::{Binary, StdError, StdResult, Uint256};
+
+use valence_encoder_utils::libraries::{
+    balancer_v2_swap::solidity_types::{multiSwapCall, swapCall},
+    updateConfigCall, Bytes32Address, ToFixedBytes,
+};
+use valence_library_utils::{msg::ExecuteMsg, LibraryAccountType};
+
+use crate::parse_address;
+
+use super::{get_update_ownership_call, get_update_processor_call};
+
+// We need to define a config and functions for this library as we don't have a CosmWasm equivalent
+#[cw_serde]
+/// Struct representing the library configuration.
+pub struct LibraryConfig {
+    /// The input address for the library.
+    pub input_addr: LibraryAccountType,
+    /// The output address for the library.
+    pub output_addr: LibraryAccountType,
+    /// The Balancer V2 Vault contract address.
+    pub vault: String,
+}
+
+#[cw_serde]
+/// Enum representing the different function messages that can be sent.
+pub enum FunctionMsgs {
+    /// Message to perform a single swap
+    Swap {
+        /// The pool ID to swap through
+        pool_id: Bytes32Address,
+        /// The address of the token to swap from
+        token_in: String,
+        /// The address of the token to swap to
+        token_out: String,
+        /// The amount to swap from the first asset
+        amount: Uint256,
+        /// The minimum amount to receive from the last asset
+        min_amount_out: Uint256,
+        /// The timeout for the swap in seconds (e.g. transaction not executed directly because it's in mempool)
+        timeout: Uint256,
+        /// Any additional data which the pool requires to perform the swap. Can be empty for all current Balancer pools.
+        /// Allows pools to have more flexible logic in the future
+        user_data: Binary,
+    },
+
+    /// Message to perform a multi-hop swap
+    MultiSwap {
+        /// List of pool IDs to swap through in order
+        /// The first pool ID is the one to swap from, the last one is the one to swap to
+        pool_ids: Vec<Bytes32Address>,
+        /// List of assets to swap through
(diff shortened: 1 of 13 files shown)


contracts/encoders/evm-encoder/src/libraries/balancer_v2_swap.rs
```diff
@@ -0,0 +1,153 @@
...
+                        .map(|data| alloy_primitives::Bytes::from(data.to_vec()))
+                        .collect::<Vec<_>>(),
+                    amount: alloy_primitives::U256::from_be_bytes(amount.to_be_bytes()),
+                    minAmountOut: alloy_primitives::U256::from_be_bytes(
+                        min_amount_out.to_be_bytes(),
+                    ),
+                    timeout: alloy_primitives::U256::from_be_bytes(timeout.to_be_bytes()),
+                };
+                Ok(multi_swap_call.abi_encode())
+            }
```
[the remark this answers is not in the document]
keyleu: this validation is done in other places

solidity/src/libraries/BalancerV2Swap.sol
```diff
@@ -0,0 +1,321 @@
...
+     * @param minAmountOut Minimum amount of output tokens to receive (slippage protection)
+     * @param timeout How long the transaction is valid for (in seconds)
+     */
+    function swap(
+        bytes32 poolId,
+        IAsset tokenIn,
+        IAsset tokenOut,
+        bytes memory userData,
+        uint256 amount,
+        uint256 minAmountOut,
```
bekauz, line 172: maybe worth to clarify `amountIn` and `amountOut`?

contracts/encoders/evm-encoder/src/libraries/balancer_v2_swap.rs
```diff
@@ -0,0 +1,153 @@
...
+        /// The pool ID to swap through
+        pool_id: Bytes32Address,
+        /// The address of the token to swap from
+        token_in: String,
+        /// The address of the token to swap to
+        token_out: String,
+        /// The amount to swap from the first token
+        amount: Uint256,
+        /// The minimum amount to receive from the last token
+        min_amount_out: Uint256,
```
bekauz, line 41: can this be disabled by setting to `0`?

bekauz, line 41: maybe worth to distinguish `amount` from `min_amount_out`, just to avoid confusion during usage/configuration

contracts/encoders/evm-encoder/src/libraries/balancer_v2_swap.rs
```diff
@@ -0,0 +1,153 @@
...
+        /// The timeout for the swap in seconds (e.g. transaction not executed directly because it's in mempool)
+        timeout: Uint256,
+        /// Any additional data which each pool requires to perform the swap. Can be empty for all current Balancer pools.
+        /// Allows pools to have more flexible logic in the future
+        user_data: Vec<Binary>,
+    },
```
bekauz, line 65: nice, I like the api

contracts/encoders/evm-encoder/src/libraries/balancer_v2_swap.rs
```diff
@@ -0,0 +1,153 @@
...
+pub enum FunctionMsgs {
+    /// Message to perform a single swap
+    Swap {
+        /// The pool ID to swap through
+        pool_id: Bytes32Address,
+        /// The address of the token to swap from
+        token_in: String,
+        /// The address of the token to swap to
+        token_out: String,
+        /// The amount to swap from the first token
```
bekauz, line 38: probably worth to add a comment that setting this to `0` will attempt to swap full input acc balance

solidity/src/libraries/BalancerV2Swap.sol
```diff
@@ -0,0 +1,321 @@
...
+
+        // Create limits array for multi-hop swap
+        // For more info on limits for batch swaps, check https://docs-v2.balancer.fi/reference/swaps/batch-swaps.html#batchswap-function
+        int256[] memory limits = new int256[](tokens.length);
+
+        // Set input token limit (maximum to spend) - Positive value indicates Balancer this is the maximum amount to spend.
+        limits[0] = int256(amountToSwap);
+
+        // Set output token limit (minimum to receive) - Negative value indicates Balancer this is the minimum amount to receive.
+        limits[tokens.length - 1] = -1 * int256(minAmountOut);
```
bekauz, line 305: interesting, shouldn't this be something like `limits[tokens.length - 1] -= int256(minAmountOut);`? right now it just overrides the limit completely which feels strange but im probably just missing context

bekauz (commented):
nice work, left a few questions regarding the solidity logic & apis we expose for this on both rust & solidity side

solidity/src/libraries/BalancerV2Swap.sol
keyleu: renamed to amountIn. Want to keep it as minAmountOut because it's not necessarily amountOut, it's the minimum you are going to receive (slippage protection)

contracts/encoders/evm-encoder/src/libraries/balancer_v2_swap.rs
keyleu: Yes, the script e2e tests with 0 here. Added this to docs.

If set to 0 it means no slippage protection though, so it's dangerous.

keyleu: Yes, renamed amount to amount_in

contracts/encoders/evm-encoder/src/libraries/balancer_v2_swap.rs
keyleu: added

solidity/src/libraries/BalancerV2Swap.sol
keyleu: the limit array is initialized above and will have zero values. So basically doing your suggested operation and mine would end up in the same result (because the current value in the array is 0).

Yours would just be doing this:
limits[tokens.length - 1] = 0 - int256(minAmountOut); 
which is the same as what I do.

I think mine is clearer though, we just want to have the negative value of -minAmountOut in that position.

Here's how the limit array works https://docs-v2.balancer.fi/reference/swaps/batch-swaps.html#batchswap-function if you want more context

bekauz: got it, thanks!

bekauz (approved):

PR #313, opened 2025-04-11
feat: vault strategist decoupling from e2e test runtime

### LINKED ISSUE #303: feat: refactor cross-chain vault strategist to operate as a background process
# description

Currently the strategist’s actions (e.g., fund transfers, entering and exiting positions, CCTP routing, etc) are invoked in the eth_vault e2e test `main()` flow in a fixed sequence. These strategist operations need to be decoupled so they can run continuously and independently (of `main()` ordering & timing) in the background.

Strategist should be spawned in a separate thread with all necessary parameters after Eth, Neutron, and Noble are set up and ready.
Any errors coming up in the strategist flow should be handled gracefully and not impact the test flow.

After this is done, `main()` should have no more notion of the strategist beyond instantiating it and starting the loop.

bekauz:
closes #303 

# Description

This is the first iteration of converting the strategist to operate as a background process.

Primary goals here were to learn about how different strategies may be implemented in Valence Protocol context, so the actual strategy logic is not perfect (e.g. the CL pool fee is hardcoded for now as it may be interpreted in more than one way).

Currently the `Strategist` deals with defining the strategy itself, as well as providing methods for it to be executed as a background process. This will change in the following PRs where all runtime/operational logic will be moved out of it and into a `ValenceService` trait.

To make testing easier, the strategist loop/epoch is configured to start at `00` seconds, every minute. Processing all strategy calls takes around 15seconds, after which strategist sleeps until the next minute. This is reflected in the e2e test flow where we await for `30` second mark every minute before simulating the user actions.

## Summary by CodeRabbit

- **New Features**
  - Added comprehensive Ethereum and Neutron account and library setups with enhanced user and contract management.
  - Introduced a strategist module for managing liquidity positions, token swaps, and cross-chain fund routing with detailed asynchronous operations.
  - Implemented Astroport decentralized exchange interactions including swap, liquidity provision, exit, and simulation functions.
  - Added utilities for precise timing controls and balance polling to optimize transaction execution and fee coverage.
  - Enhanced vault integration tests with multi-user scenarios, detailed epoch controls, and improved cross-chain fund routing.
  - Expanded contract query capabilities with new simulation and configuration response types.

- **Bug Fixes**
  - Improved logging and error handling in cross-chain fund routing and vault fee calculations.

- **Refactor**
  - Removed legacy strategist functions and vault setup utilities to streamline codebase.
  - Updated network configuration parameters for Ethereum and Neutron chains to improve cross-chain communication.
  - Modularized code structure with new strategist submodules for better maintainability.


e2e/examples/eth_vault/program.rs
```diff
@@ -544,3 +727,67 @@ pub fn setup_neutron_ibc_transfer_lib(
...
+        .get_chain(NEUTRON_CHAIN_NAME)
+        .contract_codes
+        .get(FORWARDER_NAME)
+        .unwrap();
+
+    let fwd_instantiate_msg = valence_library_utils::msg::InstantiateMsg::<
+        valence_forwarder_library::msg::LibraryConfig,
+    > {
+        owner: NEUTRON_CHAIN_ADMIN_ADDR.to_string(),
+        processor: NEUTRON_CHAIN_ADMIN_ADDR.to_string(),
```
keyleu, line 747: Are we skipping authorization/processor contract for the strategist demo/testing? I'm fine with it but I wonder where would the authorization / processor be defined if it's being used. 

Right now lite_processor is defined in EthereumProgramLibraries but it's not really a library. I guess it should go in a different section of the Strategist?

e2e/examples/eth_vault/strategist/client.rs
```diff
@@ -0,0 +1,303 @@
...
+    pub eth_program_libraries: EthereumProgramLibraries,
+
+    // underlying pool and its lp token addrress
+    pub lp_token_denom: String,
+    pub pool_addr: String,
+
+    // deposit token information on Ethereum & Cosmos
+    pub uusdc_on_neutron_denom: String,
+    pub ethereum_usdc_erc20: Address,
+}
```
keyleu, line 54: Mentioned above but here the core parts are missing, authorization(s) & processors

e2e/examples/eth_vault/strategist/routing.rs
```diff
@@ -0,0 +1,468 @@
...
+            self.neutron_client
+                .poll_for_tx(&transfer_rx.hash)
+                .await
+                .unwrap();
+        }
+    }
+
+    /// calculates the amount of shares that need to be liquidated to fulfill all
+    /// pending withdraw obligations and forwards those shares from the position
+    /// account to the withdrawal account.
```
keyleu, line 63:
```suggestion
    /// calculates the amount of shares that need to be liquidated to fulfil all
    /// pending withdraw obligations and forwards those shares from the position
    /// account to the withdrawal account.

File: e2e/examples/eth_vault/strategist/vault.rs [hunk 4]
```diff
@@ -0,0 +1,170 @@
...
+                "Calculated withdraw fee {withdraw_fee} exceeds max allowed {}, using max",
+                vault_cfg.maxWithdrawFeeBps
+            );
+            return Ok(vault_cfg.maxWithdrawFeeBps);
+        }
+
+        Ok(withdraw_fee)
+    }
+
+    async fn calculate_redemption_rate(&self) -> Result<Decimal, Box<dyn Error>> {
```
keyleu, line 62: this is kinda related to my question today. In this function how do we keep track of funds on the way (burned during CCTP transfer but not received yet). These funds might take a long time to be received but still need to be kept track of to calculate this rate. E.g. we are bridging USDC from Base to Ethereum which takes 7 days. These funds should be taken into account for calculations but they are not queriable and the strategist cant just sleep for 7 days to wait for this.

keyleu (approved):
Approving for now as most comments are not PR blocking but several things I want to confirm:

- We probably want the strategist to be interacting with the authorization contract not libraries directly. Although not required for testing we'd want to do that for the final product right (also maybe for a demo before that not sure)?

- Same for EVM, we are interacting with libraries directly but should eventually move to processor interaction and eventually lite authorizations interaction?

- Related to both points above, should we have a diff section inside Strategist for these?

- I think we might have an issue with keeping track of *under the way* tokens during bridge events or maybe I haven't grasped the full picture.

e2e/examples/eth_vault/program.rs
bekauz: yeah I kept this here for now but definitely need to adjust this to take in the actual processor. will take care of it.

e2e/examples/eth_vault/evm.rs
```diff
@@ -1,71 +1,463 @@
...
+        platformFeeBps: 10_000,    // 0.1% yearly platform fee
+        performanceFeeBps: 10_000, // 0.1% performance fee
+        solverCompletionFee: 0,    // No solver completion fee
+    };
+
+    let fee_distribution = FeeDistributionConfig {
+        strategistAccount: eth_accounts[0], // Strategist fee recipient
+        platformAccount: eth_accounts[1],   // Platform fee recipient
+        strategistRatioBps: 10_000,         // 0.1% to strategist
+    };
```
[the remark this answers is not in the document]
bekauz: cleaned this one up in here, just mate this fn accept a `VaultConfig`

e2e/examples/eth_vault/strategist/client.rs
bekauz: I added them to `NeutronProgramLibraries` and `EthereumProgramLibraries`, although not directly using them yet.

e2e/examples/eth_vault/strategist/vault.rs
bekauz: In the strategist logic we take these async steps into account by awaiting on the expected destination state.

Lets say we find that there are 100udenom in the ethereum deposit account. We then query the noble deposit account udenom balance and trigger the CCTP transfer call. Then we start polling the noble deposit account, expecting that balance to become `old_balance + 100`. Currently we go into the polling phase right after the cctp transfer call is included, but in mainnet we will likely add some constant sleep timer to not waste a bunch of queries we know would fail (e.g. if there are some finality parameters in the transport protocol which we know always apply).

As far as longer delays go, like the 7 day Base/Op -> Ethereum one you mention, the vault is expected to handle that gracefully. `VaultConfig` is configured with `withdrawLockupPeriod`, which is effectively a timeout on the withdraw requests. So for Base/Op vault this may be 7days timeout, which would start counting from the moment when the user submits a redeem request. At that point the strategist picks up the withdraw request and starts routing the funds back to the vault chain, eventually bridging the funds which will be subject to the 7days delay. Does that make sense?

keyleu: discussed it. Would only affect L2->L1, which is not the case

PR #315, opened 2025-04-14
feat: StandardBridgeTransfer library

keyleu:
Adds the StandardBridgeTransfer library which leverages the Standard Bridge (https://docs.optimism.io/app-developers/bridging/standard-bridge) to perform transfer between EVM L1s and L2s (and viceversa).

Includes: library with unit tests for validations and e2e tests for protocol interaction (a test between OP and Ethereum to test the L2->L1 interaction and a test between Base and Ethereum to test the L2->L1 interaction).

This library works for both L1StandardBridge and L2StandardBridge due to how this bridge is designed (it inherits from a StandardBridge contract) that is indirectly called by L1StandardBridge and L2StandardBridge functions, therefore we can interact with StandardBridge directly to support bridging in both directions.

Added @uditvira and @hxrts for the doc section review.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new Standard Bridge Transfer capability for secure cross-chain transfers, supporting both native ETH and ERC20 tokens.
	- Added enhanced encoding and configuration for seamless asset bridging across Layer 1 and Layer 2 networks.
	- Provided Solidity scripts to demonstrate and test transfers between L1 and L2 networks using the Standard Bridge.
  
- **Documentation**
	- Updated documentation to include comprehensive guidance on the Standard Bridge Transfer process, its configuration, and usage scenarios.
  
- **Tests**
	- Expanded test coverage to ensure correct address parsing and robust transfer functionality under various conditions.
	- Added tests to validate the behavior of the `StandardBridgeTransfer` contract, including error handling and successful transfers.

### DIFF
--- a/solidity/src/libraries/StandardBridgeTransfer.sol
+++ b/solidity/src/libraries/StandardBridgeTransfer.sol
@@ -1,0 +1,143 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.28;
+
+import {Library} from "./Library.sol";
+import {BaseAccount} from "../accounts/BaseAccount.sol";
+import {IERC20} from "forge-std/src/interfaces/IERC20.sol";
+import {IStandardBridge} from "./interfaces/standard-bridge/IStandardBridge.sol";
+
+/**
+ * @title StandardBridgeTransfer
+ * @dev Contract for automatically transferring tokens using StandardBridge on both L1 and L2.
+ * Works with both L1StandardBridge and L2StandardBridge contracts.
+ * It allows for transferring ETH or ERC20 tokens across chains.
+ */
+contract StandardBridgeTransfer is Library {
+    /**
+     * @dev Configuration struct for StandardBridge transfer parameters.
+     * @param amount The number of tokens to transfer. If set to 0, the entire balance is transferred.
+     * @param inputAccount The account from which tokens will be transferred from.
+     * @param recipient The recipient address on the destination chain.
+     * @param standardBridge The StandardBridge contract address (L1 or L2 version).
+     * @param token The ERC20 token address to transfer (or address(0) for ETH).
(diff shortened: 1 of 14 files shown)


solidity/src/libraries/StandardBridgeTransfer.sol
```diff
@@ -0,0 +1,143 @@
...
+        if (decodedConfig.inputAccount == BaseAccount(payable(address(0)))) {
+            revert("Input account can't be zero address");
+        }
+
+        // For ERC20 transfers, ensure remote token is specified
+        if (decodedConfig.token != address(0) && decodedConfig.remoteToken == address(0)) {
+            revert("Remote token must be specified for ERC20 transfers");
+        }
+
+        return decodedConfig;
```
bekauz, line 78: is there a way to validate the token/remoteToken cfg any better here? trying to wrap my head around what would happen if the remoteToken address here actually corresponds to a totally different token

bekauz: if there isn't much we can do there regarding erc20 validation, maybe we can at least check if `token` is native eth then `remoteToken` should be native eth as well (and vice versa)?

bekauz (approved):
clean, just left a few thoughts regarding the token validation in the config

solidity/src/libraries/StandardBridgeTransfer.sol
keyleu: Don't think we can do anything here, it lives in a completely different chain and can be any address (In fact an ERC20 can be represented as multiple ERC20 in the destination chain so it's not a 1 to 1, source: https://docs.optimism.io/app-developers/bridging/standard-bridge#bridged-tokens).
Regarding ETH, remoteToken is not used when you are bridging ETH, but in any case I added a check to not allow setting anything else than address(0) when you are bridging ETH, to avoid confusion.
Also extended the docs to clarify this.

Issue #314, opened 2025-04-14
feat: Valence Services trait


bekauz:
# Description

Current strategist implementation has no distinction between the actual strategy and the environment in which it operates and the way that the strategy logic is being executed.

To introduce clearer separation between the two, add a `ValenceService` trait which all strategists would implement. It should define the interface for handling various offchain services and include methods for:

- starting a service
- stopping a service
- getting the current state of the service

bekauz:
closed by #321

PR #321, opened 2025-04-17
feat: Valence workers

bekauz:
work towards #314 

# Description

This pr introduces the initial `ValenceWorker` trait which defines the api we expect various off-chain services to adhere to.

It is now implemented on both Eth Vault strategy and the mock CCTP relayer.


e2e/examples/eth_vault/strategist/strategy.rs
```diff
@@ -1,118 +1,256 @@
...
-    pub struct NeutronLibraries {
-        pub neutron_ibc_transfer: String,
-        pub noble_inbound_transfer: String,
-        pub noble_cctp_transfer: String,
-        pub astroport_lper: String,
-        pub astroport_lwer: String,
-        pub liquidation_forwarder: String,
-        pub authorizations: String,
-        pub processor: String,
+    // e2e test helper to parse the StrategyConfig from a file
```
keyleu, line 81: not really an e2e test helper right? Will be used for production as well

keyleu (approved):

e2e/examples/eth_vault/strategist/strategy.rs
bekauz: indeed, good catch!

PR #330, opened 2025-04-24
feat: anvil impersonation trait

keyleu:
Adds an AnvilImpersonationClient trait that extends EvmBaseClient to allow sending transactions from any address when running an anvil instance.

This will be useful specially for e2e tests where we want to fork networks and fund the real tokens into our accounts.

### DIFF
--- a/packages/chain-client-utils/src/ethereum.rs
+++ b/packages/chain-client-utils/src/ethereum.rs
@@ -53,6 +53,8 @@ mod tests {
         ValenceVault::{self},
     };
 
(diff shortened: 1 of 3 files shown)


bekauz (commented):
nice, looking good! wyt if we put this under a testing feature flag to avoid having this option in production strategist/worker builds?

keyleu:
> nice, looking good! wyt if we put this under a testing feature flag to avoid having this option in production strategist/worker builds?

Sure, added a feature test-utils and put anvil.rs behind it

bekauz (approved):

PR #331, opened 2025-04-24
refactor: replace etch by funding from whale in script tests

keyleu:
Remove the MockERC20 dependency in script e2e tests and instead fund from whale. Since we are changing the Mock to work with all our tests I removed this dependency in our scripts as they might break. Instead I fund from a whale account and use the real runtime code of the tokens.

### DIFF
--- a/solidity/script/BalancerV2Swap.script.sol
+++ b/solidity/script/BalancerV2Swap.script.sol
@@ -7,7 +7,6 @@ import {BalancerV2Swap} from "../src/libraries/BalancerV2Swap.sol";
 import {BaseAccount} from "../src/accounts/BaseAccount.sol";
 import {IAsset, IBalancerVault} from "../src/libraries/interfaces/balancerV2/IBalancerVault.sol";
 import {IERC20} from "forge-std/src/interfaces/IERC20.sol";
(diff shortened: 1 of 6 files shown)


solidity/script/BalancerV2Swap.script.sol
```diff
@@ -18,6 +17,9 @@ contract BalancerV2SwapScript is Script {
     address constant USDCE_ADDR = 0x549943e04f40284185054145c6E4e9568C1D3241;
     address constant BYUSD_ADDR = 0x688e72142674041f8f6Af4c808a4045cA1D6aC82;
 
+    // Whale addresses
+    address USDCE_WHALE = 0xDc927Bd56CF9DfC2e3779C7E3D6d28dA1C219969;
```
bekauz, line 21: unless it's not a typo

```suggestion
    address USDC_WHALE = 0xDc927Bd56CF9DfC2e3779C7E3D6d28dA1C219969;
```

bekauz (approved):
lgtm

solidity/script/BalancerV2Swap.script.sol
keyleu: it's not a typo, its actually called USDCe , its the one that comes from Stargate

PR #338, opened 2025-04-30
fix: allow deploying libraries with different owner than sender

keyleu:
currently libraries cant be deployed with a different owner than the sender because it calls updateConfig which has access control. Therefore when we deploy, it sets the owner to the specific address but then when calling update it fails because the sender is not the owner. This fixes it by creating a config initializer

## Summary by CodeRabbit

- **New Features**
  - Improved contract initialization by introducing dedicated internal configuration methods across multiple smart contracts. This enhances the setup process during contract deployment and ensures configuration validation at initialization.
- **Refactor**
  - Centralized and streamlined configuration logic for contract deployment, separating initial setup from public configuration updates for better maintainability.

### DIFF
--- a/solidity/src/libraries/AavePositionManager.sol
+++ b/solidity/src/libraries/AavePositionManager.sol
@@ -240,6 +240,14 @@ contract AavePositionManager is Library {
     }
 
     /**
+     * @dev Internal initialization function called during construction
+     * @param _config New configuration
+     */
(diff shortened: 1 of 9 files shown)


bekauz (approved):
Lgtm. Now that you pointed this out, it was kind of strange to be calling update config in the constructors. This is nicer.

PR #344, opened 2025-05-09
feat: mock eureka rly

### LINKED ISSUE #322: feat: mock ibc-eureka relayer
blocked by #321 

# Description

In order to enable ibc eureka-based bridging in our e2e tests, a new type of mock relayer is needed.

This should be done by making use of the Valence Worker trait (#321) and following the approach outlined by the Mock CCTP relayer (#299).

bekauz:
closes #322 

# Description

Initial implementation of a mock eureka relayer. For the first version the logic is very basic and works as follows.

Transfers from non-evm chains are routed to the mock relayer address on gaia. From there, the relayer performs two actions:
1. "burns" the funds by sending them to a dead address to avoid double counting
2. mints the equivalent amount of the specified erc20 token to the configured eth withdraw account address

Transfers from evm chains are handled by picking up on emitted eureka transfer library events, decoding them, and directly performing a tokenfactory mint on the destination chain.

This design imposes a few limitations on the way e2e flow is configured. For one, ibc-transfer libraries on neutron need to be configured in a mock mode where the eureka callback address is the addresses of relayer on gaia (while in prod this should reflect the real eureka callback contract).

In addition to that, a few changes were made to the eureka vault e2e configuration to make the testing of mock eureka relaying easier. There are still changes to be made on that front to get the vault fully functional. This will be done in a follow-up pr.

## Summary by CodeRabbit

- **New Features**
  - Introduced a mock relayer to simulate cross-chain token transfers between Ethereum and Neutron, enabling more realistic testing of IBC Eureka transfer flows.
  - Added a new client for interacting with the Cosmos Hub (Gaia) blockchain.
  - Emitted a new event for token transfers in the IBCEurekaTransfer contract for improved event tracking.

- **Bug Fixes**
  - Corrected fee configuration values to reflect accurate percentages in vault setup.

- **Refactor**
  - Improved fee handling, address reuse, and logging in routing logic for cross-chain transfers.
  - Simplified timeout and expiration handling for transfer operations.
  - Replaced dynamic values with fixed constants for certain contract addresses.
  - Removed explicit Ethereum block mining steps in favor of the new mock relayer.

- **Chores**
  - Updated configuration files with new contract addresses and account details.
  - Exposed new modules for Gaia client and mock relayer functionality.

- **Style**
  - Streamlined logging output for improved readability.

### DIFF
--- a/solidity/src/libraries/IBCEurekaTransfer.sol
+++ b/solidity/src/libraries/IBCEurekaTransfer.sol
@@ -43,11 +43,7 @@ contract IBCEurekaTransfer is Library {
      * @param _processor Address of the designated processor that can execute functions.
      * @param _config Encoded configuration parameters for the IBC Eureka transfer.
      */
-    constructor(
-        address _owner,
-        address _processor,
-        bytes memory _config
-    ) Library(_owner, _processor, _config) {}
+    constructor(address _owner, address _processor, bytes memory _config) Library(_owner, _processor, _config) {}
 
     /**
      * @dev Validates configuration by decoding the provided bytes and ensuring no critical addresses are zero.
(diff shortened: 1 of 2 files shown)


e2e/examples/eth_eureka_vault/strategist/routing.rs
```diff
@@ -287,7 +290,9 @@ impl EurekaVaultRouting for Strategy {
             "ibc/D742E8566B0B8CC8F569D950051C09CF57988A88F0E45574BFB3079D41DE6462",
             "1",
             "0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599",
-            withdraw_account_wbtc_bal.to_string(),
+            "100000",
+            // withdraw_account_wbtc_bal.to_string(), this can be too small for eureka
+            // api so hardcoding the value for now
```
keyleu, line 295: dont understand this part, does the call fail with small values?

keyleu (approved):

e2e/examples/eth_eureka_vault/strategist/routing.rs
bekauz: yeah! returns some error code that disappears when values get higher. I think this depends on the denom so it shouldn't be much of an issue, just need some reasonable defaults?

PR #353, opened 2025-05-21
feat: One way vault

keyleu:
New version of vault focused only on depositing assets and storing the withdraw requests in the contract state by their ID (incrementing value).

The purpose of this is to eventually prove these using the ZK coprocessor and generate proof for the withdraw requests that will be passed to the corresponding contracts on the other domain (in this case Neutron).

Took the chance to reorganize the folders a little bit and moved vault contracts under vaults instead of libraries.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new upgradeable, pausable vault supporting one-way asset deposits, withdrawal requests across chains, and ERC4626 compliance.
  - Added configurable deposit fees, deposit caps, and fee distribution between platform and strategist.
  - Enabled strategist-controlled redemption rate updates and advanced access controls for vault operations.
  - Implemented pausing/unpausing functionality with state tracking.

- **Tests**
  - Added comprehensive tests for the new vault covering initialization, configuration, deposits, withdrawals, fee handling, access control, upgradeability, and lifecycle scenarios.
  - Included fuzz and stress testing for robustness.

- **Chores**
  - Updated import paths in existing test files for improved directory structure; no changes to test logic or behavior.


solidity/src/vaults/OneWayVault.sol
```diff
@@ -0,0 +1,672 @@
...
+
+        uint256 totalDeposits = totalAssets();
+        if (totalDeposits >= cap) {
+            return 0;
+        }
+
+        unchecked {
+            return cap - totalDeposits;
+        }
+    }
```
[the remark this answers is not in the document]
keyleu: this is fine, the feesOwed should not be included in the deposit cap

solidity/src/vaults/OneWayVault.sol
```diff
@@ -0,0 +1,672 @@
...
+        emit ConfigUpdated(msg.sender, decodedConfig);
+    }
+
+    /**
+     * @dev Validates the configuration parameters
+     * @param decodedConfig The decoded OneWayVaultConfig struct
+     */
+    function _validateConfig(OneWayVaultConfig memory decodedConfig) internal pure {
+        if (address(decodedConfig.depositAccount) == address(0)) {
+            revert("Deposit account cannot be zero address");
```
bekauz, line 263: can we define this and other reverts as dedicated error types?

solidity/src/vaults/OneWayVault.sol
```diff
@@ -0,0 +1,672 @@
...
+        uint256 grossAssets = baseAssets.mulDiv(BASIS_POINTS, BASIS_POINTS - feeBps, Math.Rounding.Ceil);
+
+        uint256 fee;
+        unchecked {
+            fee = grossAssets - baseAssets;
+        }
+
+        if (fee > type(uint256).max) {
+            revert("Fee exceeds uint256 max");
+        }
```
bekauz, line 474: can this evaluate to true?

solidity/src/vaults/OneWayVault.sol
```diff
@@ -0,0 +1,672 @@
...
+    function calculateDepositFee(uint256 assets) public view returns (uint256) {
+        uint32 feeBps = config.depositFeeBps;
+        if (feeBps == 0) return 0;
+
+        uint256 fee = assets.mulDiv(feeBps, BASIS_POINTS, Math.Rounding.Ceil);
+
+        // Check if fee exceeds uint256 max value
+        if (fee > type(uint256).max) {
+            revert("Fee exceeds uint256 max");
+        }
```
bekauz, line 496: same as above

solidity/src/vaults/OneWayVault.sol
```diff
@@ -0,0 +1,672 @@
...
+        __ERC4626_init(IERC20(underlying));
+        __Ownable_init(_owner);
+        __ReentrancyGuard_init();
+
+        config = abi.decode(_config, (OneWayVaultConfig));
+        _validateConfig(config);
+
+        unchecked {
+            ONE_SHARE = 10 ** decimals();
+            redemptionRate = startingRate; // Initialize at specified starting rate
```
bekauz, line 228: does it make sense to have this assignment in `unchecked` scope?

solidity/src/vaults/OneWayVault.sol
```diff
@@ -0,0 +1,672 @@
...
+    function _distributeFees(FeeDistributionConfig memory feeDistribution) internal {
+        uint256 _feesOwedInAsset = feesOwedInAsset;
+        if (_feesOwedInAsset == 0) return;
+
+        // Calculate fee shares for strategist
+        uint256 strategistAssets =
+            _feesOwedInAsset.mulDiv(feeDistribution.strategistRatioBps, BASIS_POINTS, Math.Rounding.Floor);
+
+        // Calculate platform's share as the remainder
+        uint256 platformAssets = _feesOwedInAsset - strategistAssets;
```
bekauz, line 631: this can be `unchecked` right ?

keyleu:
> first stab, left a couple of questions and suggestions.
> 
> overall I enjoy the solidity pattern where all errors are defined at the beginning of a contract, maybe it would be nice here as well?

reason why I use strings is because they are easily decoded, and we forward the execution results from processor to authorizations contract. This is why I use strings in all our reverts. If we ever hook this to the authorizations contract + processor, we can easily decode the execution results that we store in the authorizations contract without knowing into what we need to abi decode them.

bekauz:
oh, that makes sense. is there no solidity/forge/etc pattern for dealing with something like this? sounds like others must have faced this before

solidity/src/vaults/OneWayVault.sol
```diff
@@ -0,0 +1,672 @@
...
+     * @param strategist Address of the vault strategist
+     * @param depositFeeBps Fee charged on deposits in basis points (1 BPS = 0.01%)
+     * @param depositCap Maximum assets that can be deposited (0 means no cap)
+     * @param feeDistribution Configuration for fee distribution between platform and strategist
+     */
+    struct OneWayVaultConfig {
+        BaseAccount depositAccount;
+        address strategist;
+        uint32 depositFeeBps;
+        uint128 depositCap;
```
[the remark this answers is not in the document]
keyleu: I guess to save storage slots. Took this from previous vault. Changed it to uint256

solidity/src/vaults/OneWayVault.sol
```diff
@@ -0,0 +1,672 @@
...
+    mapping(uint64 => WithdrawRequest) public withdrawRequests;
+
+    // Constants
+
+    /**
+     * @dev Constant for basis point calculations (100% = 10000)
+     */
+    uint32 private constant BASIS_POINTS = 1e4;
+
+    uint256 internal ONE_SHARE;
```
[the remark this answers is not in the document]
keyleu: nope, this depends on decimals() which is defined in init function in the initializer. It's only initialized there though.

solidity/src/vaults/OneWayVault.sol
keyleu: explained below

solidity/src/vaults/OneWayVault.sol
keyleu: cant. Removed it because now they are both uint256

solidity/src/vaults/OneWayVault.sol
keyleu: removed all uncheckeds, they are not needed and they save barely anything (40 gas), prefer better readability

solidity/src/vaults/OneWayVault.sol
keyleu: it could, I removed all uncheckeds because the gas we are saving with them is minimal, basically nothing. So I'm going for a better readability here.

It would make sense for a loop I guess where we are doing many operations, but for these single subtractions it's a bit meh

solidity/src/vaults/OneWayVault.sol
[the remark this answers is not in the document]
bekauz: I think immutable would work then, just read about it in [the docs](https://docs.soliditylang.org/en/latest/contracts.html#immutable) myself

bekauz (approved):

solidity/src/vaults/OneWayVault.sol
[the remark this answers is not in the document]
keyleu: Gotta test it, will do it in the next PR. Thing is we are not using the constructor because of the initializing/proxy way.

PR #357, opened 2025-05-23
feat: clearing library

### LINKED ISSUE #350: feat: clearing house library
# Description

A new type of clearing smart contract is needed to perform the following actions:

- accept incoming withdrawal obligations and store them in a FIFO queue
- allow an authorized strategist to settle the oldest items in the queue

which should be enabled with the following execute message:

```rust
pub enum ExecuteMsg {
    /// validates and enqueues a new withdrawal obligation
    RegisterObligation(WithdrawalObligation),
    /// settles the oldest withdrawal obligation
    SettleNextObligation {},
}
```

## Withdrawal obligation registration

Example definition may look as follows:

```rust
/// unsettled liability sitting in the clearing queue
WithdrawalObligation {
    recipient: String,       // where the payout is to be routed
    payout_coins: Vec<Coin>  // what is owed to the recipient
}
```

Withdrawal obligations that arrive to this contract are assumed to be correct due to some
actions on other domains.

The contract should ensure that:
1.  the withdrawal obligation being registered is coming from a valid source
2. the recipient address is valid

If both checks pass, the obligation is pushed to the queue.

## Settlement account

The settlement account associated with this contract is a [Valence Base account](https://docs.valence.zone/accounts/base_accounts.html).

In Valence Protocol terms, settlement account will function as the input account from which
funds are to be routed to the final recipients.

**Funding of the settlement account is outside the scope of this contract**. This contract is 
just the settlement engine.

Liquidity-management flow, which will also be operated by the strategist, will be responsible
for continuously monitoring the settlement account balance and the obligation queue in order
to keep the settlement account topped up with sufficient/desired liquidity.

## Obligation settlement

Obligations are settled by the approved strategist.

Settlement happens by transferring funds from the settlement account to the recipient specified
in each `WithdrawalObligation` being processed.

This contract will delegate `BankSend` instructions to the settlement account in order to
 carry out the final payouts.

Settlement should happen as follows:

1. pop the head of the obligation queue and error out if it's empty
2. validate that the associated settlement (input) account has sufficient funds for the payout
3. delegate the payout transfer message to the settlement account 

# Future work

In the future, this obligation registration and settlement can be batched. Keeping it simple for now.

bekauz:
closes #350 

# Description

Introduces the clearing queue library.

This library allows registering and settling obligations.

## Sender validation

Any message sender validations are expected to be done on authorizations level.

## Obligation registration

Obligations are registered with the following execute msg variant:

```rust
/// validates and enqueues a new withdrawal obligation
RegisterObligation {
    /// where the payout is to be routed
    recipient: String,
    /// what is owed to the recipient
    payout_coins: Vec<Coin>,
    /// some unique identifier for the request
    id: Uint64,
},
```

In turn, the following validations are performed:
1. obligation `id` was not registered before
2. `payout_coins` vector is not empty
3. every coin in `payout_coins` amount is non-zero
4. `recipient` address is valid

Following the validations, the constructed obligation is pushed to the end of the queue,
**and** store the obligation `id` to prevent double accounting.

From here, the validation is considered valid and ready to be settled.

## Obligation settlement

Obligations are settled with the following execute msg variant:

```rust
/// settles the oldest withdrawal obligation
SettleNextObligation {},
```

In turn, the following validations are performed:
1. obligation queue has at least one element
2. settlement input account has sufficient funds for settling every coin to be paid out

After that, a `BankSend` message is constructed and delegated to be executed on behalf
of the settlement (input) account.


contracts/libraries/clearing-queue/src/state.rs
```diff
@@ -0,0 +1,28 @@
...
+pub struct WithdrawalObligation {
+    /// where the payout is to be routed
+    pub recipient: Addr,
+    /// what is owed to the recipient
+    pub payout_coins: Vec<Coin>,
+    /// some unique identifier for the request
+    pub id: Uint64,
+    /// block when registration was enqueued
+    pub enque_block: BlockInfo,
+}
```
keyleu, line 28: would also add the request_addr here (address who requested on Ethereum in this case) and the status (pending/completed)

contracts/libraries/clearing-queue/src/state.rs
```diff
@@ -0,0 +1,28 @@
+use cosmwasm_schema::cw_serde;
+use cosmwasm_std::{Addr, BlockInfo, Coin, Empty, Uint64};
+use cw_storage_plus::Map;
+use valence_processor_utils::queue::QueueMap;
+
+/// set of registered obligation ids meant to prevent double filling
+/// of any given obligation
+pub const REGISTERED_OBLIGATION_IDS: Map<u64, Empty> = Map::new("reg_obl_id");
```
keyleu, line 8: So currently we can only know what withdrawal obligations are currently being processed (currently in the queue), and these are currently not queriable by address.

I suggest changing this `REGISTERED_OBLIGATIONS_IDS` into an IndexedMap with a MultiIndex (request_addr, which would need to be added to WithdrawalObligation, along with the status, which would need to be updated after being processed). 
This way we can query:
- Withdrawal obligation for a specific ID.
- Withdrawal obligations per request address. For this query we would get all withdrawal obligations that an eth address requested along with their current status.

This would be enough in case the front end would like to display this information for a user. Otherwise we don't really know much of what happened to withdrawal obligations because once they are processed they are just gone

LMK what you think.

keyleu (commented):
Current implementation works, left some suggestions for state refactor that would allow more queries to know what's going on and what happened.

contracts/libraries/clearing-queue/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Configuration
+
+The library is configured on instantiation via the `LibraryConfig` type.
+
+```rust
+pub struct LibraryConfig {
+    /// settlement input account which we tap into in order
+    /// to settle the obligations
+    pub input_addr: LibraryAccountType,
```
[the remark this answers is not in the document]
bekauz: oh yeah I forgot to change into this, done now

contracts/libraries/clearing-queue/src/state.rs
bekauz: this kind of follows from my request above.

happy to jam on this if you feel like it's worthwhile, but in the current design `request_addr` would be purely informational and useful only for frontend as it has no impact on the settlements

contracts/libraries/clearing-queue/src/state.rs
bekauz: my thinking was that the frontend already has state information about both eth & neutron.

in this case, the frontend will know the user address and the withdraw requests created by that user on the OneWayVault on eth. from those withdraw requests we know their unique ids.

to get each withdraw request status, fe queries the entire neutron obligation queue and filters it by the withdraw request id. if filtering gives a result, withdraw is in progress. otherwise it's been completed.

does that make sense?

contracts/libraries/clearing-queue/src/msg.rs
```diff
@@ -0,0 +1,115 @@
...
+    QueueInfo {},
+    /// returns a list of obligations in the queue starting from the given index
+    #[returns(ObligationsResponse)]
+    Obligations {
+        /// starting index
+        from: Option<u64>,
+        /// end index
+        to: Option<u64>,
+    },
+}
```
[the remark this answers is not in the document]
bekauz: I'm not sure if strategist should be touching neutron for proof generation. feels a bit fragile given the latency of the coprocessor?

contracts/libraries/clearing-queue/src/state.rs
keyleu: yeah, let's talk about this. The minimum we need here is a query on this Map here for the Front end to know if something is pending to be registered or not. Otherwise we can't know if something hasn't been registered yet or is completed. That's the bare minimum, rest would just be extras to make it easier

contracts/libraries/clearing-queue/src/msg.rs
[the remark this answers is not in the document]
keyleu: OK let's say the strategist stops working and we restart it.
How does it know what current withdraw request it needs to register?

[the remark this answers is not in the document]
keyleu: Let's say the clearing queue is empty. And the strategist restarts. How does it know which ID to use to start asking for proves for?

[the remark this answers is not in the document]
bekauz: yeah, I think this problem applies not just in situations where strategist restarts but also in cases where withdraw obligations get posted out of order.

tbh I don't have a good answer because I don't know the internals of the coprocessor. hence the decoupled design of this contract - I assumed that coprocessor + strategist + ethereum state is sufficient to shield the settlement flow from the way those requests originate (other than the withdraw obligation id checks).  

cc @hxrts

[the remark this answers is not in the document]
keyleu: Coprocessor has no state. Will only proof that a certain ID exists and give the necessary information to this contract.
We can make the strategist post the obligations always in order, but for that it needs to at least know the last one posted, currently we have no way of knowing the last posted because the clearing queue might be empty and we have no access to the other map via a query

contracts/libraries/clearing-queue/src/state.rs
bekauz: on a second thought, what if instead of 

```rust
pub const REGISTERED_OBLIGATION_IDS: Map<u64, Empty> = Map::new("reg_obl_id");
```

we had 

```rust
pub const REGISTERED_OBLIGATION_IDS: Map<u64, bool> = Map::new("reg_obl_id");
```

where `bool` indicates whether the request is completed?

keyleu: Yes, that solves it too, was going to eventually suggest this. It's basically similar to my solution but without adding the request_addr from ethereum, which is just, like you said, informational (and potentially useful for the fe to q withdraw_requests per request_addr).

bekauz: added the query + updated the obligations map to store the status as the value

PR #360, opened 2025-05-28
zk coprocessor docs update

hxrts:
this reformulates the intro + SMT docs that were on the docs site, and then adds a few more documents that fill in details about some important components:
- the coprocessor service architecture
- how the zk system hooks into the on-chain contracts
- a developer guide for using the app template

this will definitely need to be adjusted as the architecture continues to solidity, but it's a much better starting point than what's on the site right now

## Summary by CodeRabbit

- **Documentation**
  - Added comprehensive new documentation covering the Valence ZK system architecture, developer guides, on-chain integration, coprocessor internals, sparse Merkle trees, guest environment, and state encoding and encoders.
  - Updated documentation structure and summary to include detailed subsections and new examples.
  - Revised introductory and overview materials for improved clarity and alignment with the Valence ZK System.
  - Removed outdated documentation related to the previous ZK coprocessor and sparse Merkle trees to streamline content.


keyleu (approved):
makes sense to me, left 1 suggestion

PR #368, opened 2025-05-31
feat: clearing queue v2

### LINKED ISSUE #366: feat: update `ClearingQueue` library
# Description

A few changes are needed to the clearing queue library (#357).

## Obligation ordering

First, the library should enforce a strict ordering of the withdrawal obligations being registered based on the `id` field.

In addition to that, a new query should be exposed that returns the latest registered withdrawal obligation id.

## Single denom

The clearing queue should operate in a single denom mode.
The denom should be configured during the instantiation.

As a result, the `RegisterObligation` function can be simplified:

```rust
    /// validates and enqueues a new withdrawal obligation
    RegisterObligation {
        /// where the payout is to be routed
        recipient: String,
        /// what is owed to the recipient
        payout_coins: Vec<Coin>, // this can become payout_amount: Uint128
        /// some unique identifier for the request
        id: Uint64,
    },
```

During the settlement requests, the library should construct the coin to be paid out by looking at a given obligation `payout_amount` and applying the configured denom to that amount.

bekauz:
closes #366 

# Description

Updates the clearing queue library.

Main changes are as follows:

- library `Config` now stores two new fields: `denom: String` to identify the clearing denom, and `latest_id: Uint64` to track the latest registered obligation id
- `LibraryConfig` takes in a `latest_id: Option<Uint64>`: if set to `None`, the obligation queue starts processing withdraw obligations from id 0. Otherwise, it allows to configure a "starting point" id and beings incrementing it from there.
- `FunctionMsgs::RegisterObligation` now accepts `payout_amount: Uint128` instead of a vector of coins to match the coprocessor capabilities

## Summary by CodeRabbit

- **New Features**
  - Added support for configuring a base denomination and an optional starting obligation ID for the clearing queue.
  - Obligations are now processed in strict, monotonically increasing order of IDs to prevent out-of-order errors.

- **Refactor**
  - Simplified obligation payouts to use a single amount in a fixed denomination instead of multiple coins.
  - Introduced richer obligation status tracking with explicit states: InQueue, Processed, and Error.
  - Updated related configuration, message structures, state management, and testing suite to reflect these changes.

- **Documentation**
  - Enhanced documentation to clarify obligation ordering, denomination configuration, and updated function parameters.

### DIFF
--- a/contracts/libraries/clearing-queue/src/contract.rs
+++ b/contracts/libraries/clearing-queue/src/contract.rs
@@ -1,23 +1,26 @@
 #[cfg(not(feature = "library"))]
 use cosmwasm_std::entry_point;
 
-use cosmwasm_std::{to_json_binary, Binary, Deps, DepsMut, Env, MessageInfo, Response, StdResult};
+use cosmwasm_std::{
+    to_json_binary, Binary, Deps, DepsMut, Env, MessageInfo, Reply, Response, StdResult,
+    SubMsgResult,
+};
 use valence_library_utils::{
     error::LibraryError,
     msg::{ExecuteMsg, InstantiateMsg},
 };
 
 use crate::{
-    msg::{
-        Config, FunctionMsgs, LibraryConfig, LibraryConfigUpdate, ObligationStatusResponse,
-        QueryMsg,
-    },
-    state::REGISTERED_OBLIGATION_IDS,
+    msg::{Config, FunctionMsgs, LibraryConfig, LibraryConfigUpdate, QueryMsg},
+    state::{ObligationStatus, OBLIGATION_ID_TO_STATUS_MAP},
 };
 
 // version info for migration info
 const CONTRACT_NAME: &str = env!("CARGO_PKG_NAME");
 const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
+
+// reply ID for invalid obligation handling
+// const OBLIGATION_SETTLEMENT_REPLY_ID: u64 = 2;
 
 #[cfg_attr(not(feature = "library"), entry_point)]
 pub fn instantiate(
@@ -64,14 +67,16 @@ mod execute {
 
 mod functions {
     use cosmwasm_std::{
-        ensure, BankMsg, Coin, DepsMut, Env, MessageInfo, Response, Uint128, Uint64,
+        ensure, BankMsg, Coin, DepsMut, Env, MessageInfo, Response, SubMsg, Uint128, Uint64,
     };
 
     use valence_library_utils::{error::LibraryError, execute_on_behalf_of};
 
     use crate::{
         msg::{Config, FunctionMsgs},
-        state::{WithdrawalObligation, CLEARING_QUEUE, REGISTERED_OBLIGATION_IDS},
+        state::{
+            ObligationStatus, WithdrawalObligation, CLEARING_QUEUE, OBLIGATION_ID_TO_STATUS_MAP,
+        },
     };
 
     pub fn process_function(
@@ -114,16 +119,10 @@ mod functions {
             ))
         );
 
-        // obligation payouts cannot be empty
-        ensure!(
-            !payout_amount.is_zero(),
-            LibraryError::ExecutionError(
-                "obligation amount must be greater than 0 in order to be registered".to_string()
-            )
-        );
-
         let withdraw_obligation = WithdrawalObligation {
-            recipient: deps.api.addr_validate(&recipient)?,
+            // intentionally not verifying the recipient address here as
+            // we do not want to block the queue processing
+            recipient,
             payout_coin: Coin {
                 amount: payout_amount,
                 denom: cfg.denom.to_string(),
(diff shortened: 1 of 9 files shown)


contracts/libraries/clearing-queue/src/contract.rs
```diff
@@ -82,51 +84,50 @@ mod functions {
...
-        payout_coins: Vec<Coin>,
+        payout_amount: Uint128,
         id: Uint64,
     ) -> Result<Response, LibraryError> {
-        // validate that this obligation is not registered yet
+        // increment the latest obligation id to the expected value
+        cfg.latest_id = cfg
+            .latest_id
+            .checked_add(Uint64::one())
+            .map_err(|_| LibraryError::ExecutionError("id overflow".to_string()))?;
```
keyleu, line 106: There is a problem with this v2 library. We might not be able to register an obligation. And there's 2 reason that might happen:
- Someone put in an invalid neutron address as recipient.
- The payout amount is zero (would be silly but imagine a vault with 100% withdraw fee, or some small amount that gets rounded to 0). 

Anyways just because of the first possibility we can't do this.
So we have 2 options: don't validate order, or validate order differently (checking that the ID requested is higher than last one, not +1).
This might surface another problem. What if the recipient registers an obligation with a certain ID but he incorrectly skipped one (timeout of coprocessor), then that previous one won't be able to be registered anymore and funds for that user can't be withdrawn.

My suggestion is to not do any order checks in this library

bekauz: yeah, good catch. regarding invalid neutron address, what if we did a soft bech32 validation on the vault contract? and as far as the payout amount goes, I'm honestly not sure what's the correct solution.

in both cases above there must be some validation or corner-cutting in order to submit a successful banksend message (which would fail if the addr is invalid or the amount is 0).

I do think that this library needs to enforce the order, though. Doing it this way makes the lifecycle of this contract manageable. Consider a situation where a strategist needs to restart - at that point we would need to do some complex querying of the clearing library and the vault events, and start matching them one by one. This is fine if we had 10 withdraw events in total, but if we get to 1000s, it will get extremely messy. On the other hand, if we are processing events in order,  migrating to a new clearing queue or new strategist is easy - we know the last processed withdraw request id and can just start from that point.
To add to that, I think it would be wrong to process withdraw requests in a non-fifo manner. If something is going wrong and everyone is rushing to exit, I would expect the settlement logic to respect the order of user withdraw requests.

keyleu: We can do a soft validation but don't think we can do a full validation of the address (we probably also don't want to make the contract only for neutron but for any destination chain). I don't think that we can assume all obligations will necessary be "valid".

We can still enforce order by storing the last registered obligation. If there is a restart we can restart by that ID. Nevertheless, if are only processing in order, the strategist should make sure it doesn't skip any valid IDs. The easiest solution would be to store the invalid ones as completed or invalid then

bekauz: @keyleu updated the settlement design, lmk what you think

contracts/libraries/clearing-queue/src/contract.rs
```diff
@@ -1,24 +1,27 @@
...
+    msg::{Config, FunctionMsgs, LibraryConfig, LibraryConfigUpdate, QueryMsg},
+    state::{ObligationStatus, OBLIGATION_ID_TO_STATUS_MAP},
 };
 
 // version info for migration info
 const CONTRACT_NAME: &str = env!("CARGO_PKG_NAME");
 const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
 
+// reply ID for invalid obligation handling
+// const OBLIGATION_SETTLEMENT_REPLY_ID: u64 = 2;
```
keyleu, line 23: leftover

contracts/libraries/clearing-queue/src/contract.rs
```diff
@@ -82,51 +89,44 @@ mod functions {
...
         let withdraw_obligation = WithdrawalObligation {
-            recipient: deps.api.addr_validate(&recipient)?,
-            payout_coins,
+            // intentionally not verifying the recipient address here as
+            // we do not want to block the queue processing
+            recipient,
+            payout_coin: Coin {
+                amount: payout_amount,
+                denom: cfg.denom.to_string(),
+            },
```
keyleu, line 129: why can't we do the checks here already? Then we can get better errors in the status rather than some cosmos-sdk invalid coins error or wasm code 5. What I mean is, do the validation of address here and that payout_amount is not zero. If neither of these conditions happen we follow the normal flow, otherwise we don't bother adding to the queue and immediately store the Obligation_ID_To_status as error.

This way we don't need the retry entry point and won't fire any invalid messages of input account. I feel code would be much cleaner

bekauz: yeah at first I liked how reply was a flexible way to catch all errors that may come up but after a second thought I think you're right. switched to a reply-less design, lmk what you think

keyleu (approved):

PR #370, opened 2025-06-02
Implementing Elys lending library

stana-miric:
This PR adds implementation for Elys lending library.

*NOTE: since the contracts are not yet deployed on Elys network (permissioned cosmwasm), the library is not tested.

### DIFF
--- a/contracts/libraries/elys-lending/src/msg.rs
+++ b/contracts/libraries/elys-lending/src/msg.rs
@@ -1,0 +1,116 @@
+use cosmwasm_schema::{cw_serde, QueryResponses};
+use cosmwasm_std::{Addr, Deps, DepsMut, Uint128, Uint64};
+use cw_ownable::cw_ownable_query;
+use valence_lending_utils::elys::ElysQuery;
+use valence_library_utils::LibraryAccountType;
+use valence_library_utils::{error::LibraryError, msg::LibraryConfigValidation};
+use valence_macros::{valence_library_query, ValenceLibraryInterface};
+
+#[cw_serde]
+pub enum FunctionMsgs {
+    /// Message to lend tokens.
+    Lend {},
+    /// Message to withdraw tokens. If amount is not specified, full amount will be withdrawn.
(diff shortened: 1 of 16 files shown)


contracts/libraries/elys-lending/src/msg.rs
```diff
@@ -0,0 +1,116 @@
...
+
+#[cw_serde]
+pub enum FunctionMsgs {
+    /// Message to lend tokens.
+    Lend {},
+    /// Message to withdraw tokens. If amount is not specified, full amount will be withdrawn.
+    Withdraw { amount: Option<Uint128> },
+    /// Message to claim reward tokens.
+    ClaimRewards {},
+}
```
keyleu, line 17: Are rewards not automatically claimed when unbonding? 
Depending on the case I see a possible improvement that can be done here:

- If not automatically claimed, when we withdraw, we also claim rewards and send them to the output account.
- If automatically claimed, when we withdraw we need to send the rewards claimed as well otherwise they will be stuck in the input account

keyleu (commented):
Let's wait to merge until tested on Elys, hopefully soon.

Only have a question regarding rewards as normally for most protocols the rewards are automatically claimed when unbonding, is this not the case for Elys?

contracts/libraries/elys-lending/src/msg.rs
stana-miric: Unfortunately, the rewards aren’t claimed automatically, although they mentioned in the Telegram group that they would fix it so they get claimed by default. I could ask if that has been resolved.
Either way, I agree that it would be nicer if claiming happened automatically when withdrawing. I’m just not sure whether it’s still better to keep the claim as a separate function or remove it entirely?

keyleu: Hm, I can see maybe we want to keep it in case we want to claim rewards and keep lending? I'm fine with keeping it since you already wrote it.

Yeah if they are going to change claim rewards we probably have to wait until they change it and then adjust the library to send the claimed rewards on withdraw.

PR #379, opened 2025-06-04
Restrict receiver to 46 or 66 characters and store flag in contract

keyleu:
This is only for the neutron deployment


bekauz (approved):
lgtm

PR #386, opened 2025-06-09
feat: update clearing lib

bekauz:


contracts/libraries/clearing-queue-supervaults/src/contract.rs
```diff
@@ -120,123 +141,57 @@ mod functions {
...
+            )
+            .map_err(|_| {
+                LibraryError::ExecutionError("failed to apply settlement ratio".to_string())
+            })?;
+
+        let supervaults_amount = match payout_amount.checked_sub(mars_amount) {
+            Ok(amt) => amt,
+            // in case payout amount was 0, this may panic. we swallow the error and mark the
+            // obligation as processed before returning.
+            Err(e) => return swallow_obligation_registration_err(deps, id, e.to_string()),
```
keyleu, line 175: This here is unreachable. Since payout amount zero is already checked before, this error will never happen. We either can specify this in the comment or use a saturating_sub for example

contracts/libraries/clearing-queue-supervaults/src/contract.rs
```diff
@@ -254,38 +209,44 @@ mod functions {
...
             // to not block subsequent withdraw obligations from being registered.
-            Err(e) => {
-                OBLIGATION_ID_TO_STATUS_MAP.save(
-                    deps.storage,
-                    id.u64(),
-                    &ObligationStatus::Error(e.to_string()),
-                )?;
-                return Ok(Response::default());
-            }
+            Err(e) => return swallow_obligation_registration_err(deps, id, e.to_string()),
```
keyleu, line 212: I don't think we should do this. If the query fails, something is being done really really wrong (maybe the API of the contract did a breaking change for example). If this is done we'd just start marking all withdraws as completed. I think if this query fails we should error for real and investigate what's happening

PR #394, opened 2025-06-12
feat: Mars Borrowing and Repayment Integration

jonas089:
This PR adds **borrow** and **repay** functionality to the Mars lending library.
In order to verify that the methods were implemented correctly, I deployed the library contract to neutron mainnet. Then I approved it as a library on a Valence base account, called **lend** on said account to create & fund the credit account and then tested the new borrow & repay endpoints:

Here the mars library contract: [neutron1ecutnjnqqrge0m8gxdx33qyddz4v64ys4fs6hlfn5uyvkwdj9mrsccct3s](https://neutron.celat.one/neutron-1/contracts/neutron1ec7r63vmdyvyvdx8frsyjtd3yn0z7xnr5umnc0quf09wapc6ty0saxwpxr)
Here the base account I used: [neutron12a2xmm5r5gk3s33kp46qz7t73h8d0d57ttvmstqfqvc50nzlqm6qe08c8n](https://neutron.celat.one/neutron-1/contracts/neutron12a2xmm5r5gk3s33kp46qz7t73h8d0d57ttvmstqfqvc50nzlqm6qe08c8n)

Here the tx hash of the borrow call:
https://neutron.celat.one/neutron-1/txs/10CBB76C7A6036984EE0D6B4AA82D34EA5FE934590284C17845FA789E4AFA1B3

Here the tx hash of the repay call:
https://neutron.celat.one/neutron-1/txs/AF6FD2D59D56BE74DF5292E2841255F422AB7590CE02A67B2ED1C72FFE3ADB74

## Summary by CodeRabbit

- **New Features**
  - Introduced the ability to borrow and repay assets directly through the Mars Lending interface.
- **Documentation**
  - Updated documentation to describe new borrowing and repayment functionalities, including detailed process explanations.

### DIFF
--- a/docs/src/libraries/cosmwasm/mars_lending.md
+++ b/docs/src/libraries/cosmwasm/mars_lending.md
@@ -90,12 +90,10 @@ The library automatically handles Mars credit account lifecycle:
 2. **Borrow Execution**: Executes the borrow action through the Mars credit account, which:
    - Borrows the specified amount of the given denom
    - Transfers the borrowed tokens to the output account specified in the library configuration
-3. **Token Transfer**: The borrowed tokens are automatically transferred to the output account
 
 ### Repayment Process
 
-2. **Token Transfer**: Transfers the repayment tokens from the input account back to the credit account
-3. **Repay Execution**: Executes the repay action through the Mars credit account
+1. **Token Transfer**: Transfers the repayment tokens from the input account back to the credit account
 
 ### Error Handling
 


keyleu (commented):
Even though the Borrow and Repay are correct, it's not 100% what we want because it doesn't really work like described in the docs you added, there's a slight difference.
Currently the borrow doesn't borrow to the output account but instead leaves the tokens in the credit account. Then the repay repays from the credit account as well. So both functions work but we might want something more useful (like the functionality you described in the docs).
Let's do this:
- On the borrow function, add a WithdrawToWallet action so that right after borrowing, we put the borrowing amount in the output account (similar to what we have in the Withdraw message). This will provide the functionality you describe in the docs, borrow and deposit it in output account (currently it goes to credit account).

- In the repay function, let's replace the Repay action for RepayFromWallet action(this will slightly change the message and require adding the funds in the funds part of the CosmwasmExecuteMsg. You can leave the API of the library as it currently is (Repay) but the action we do on Mars will be RepayFromWallet. 
Sorry for telling you to use repay instead of repayfromwallet but it was because I wasn't that familiar with the Mars + Credit account functionality, now after reading the docs I see how it works.

In the end this will give us the functionality we want: borrow something and deposit it somewhere so we can use it for other things, and when we want to repay we just bring the amount to repay to the input account and trigger the Repay message.

docs/src/libraries/cosmwasm/mars_lending.md
```diff
@@ -82,6 +84,19 @@ The library automatically handles Mars credit account lifecycle:
...
+1. **Credit Account Check**: Verifies the existence of a credit account for the input address
+2. **Borrow Execution**: Executes the borrow action through the Mars credit account, which:
+   - Borrows the specified amount of the given denom
+   - Transfers the borrowed tokens to the output account specified in the library configuration
+3. **Token Transfer**: The borrowed tokens are automatically transferred to the output account
+
+### Repayment Process
+
+2. **Token Transfer**: Transfers the repayment tokens from the output account (or any other account) to the input account
+3. **Repay Execution**: Executes the repay action through the Mars credit account
```
keyleu, line 98: This one is a bit different. There's no token transfer.
It´s a single step that repays the tokens specified from the input account balance

docs/src/libraries/cosmwasm/mars_lending.md
```diff
@@ -39,6 +39,8 @@ graph LR
 |----------|------------|-------------|
 | **Lend** | - | Creates a Mars credit account (if one doesn't exist) and lends the entire balance of the specified denom from the input account to the Mars Protocol through the credit account. |
 | **Withdraw** | `amount: Option<Uint128>` | Withdraws lent assets from the Mars credit account to the output account. If no amount is specified, withdraws the entire position. |
+| **Borrow** | `coin: Coin` | Borrows the specified amount of the given denom from Mars Protocol through the credit account. The borrowed tokens are sent to the output account specified in the library configuration. |
+| **Repay** | `account_id: String`, `coin: Coin` | Repays borrowed assets to Mars Protocol through the input account. The `Coin` parameter contains `denom` and `amount` fields. |
```
keyleu, line 43: can remove account_id from here

docs/src/libraries/cosmwasm/mars_lending.md
```diff
@@ -82,6 +84,19 @@ The library automatically handles Mars credit account lifecycle:
    - `Reclaim`: Withdraws the lent position back to the credit account
    - `WithdrawToWallet`: Transfers the tokens from credit account to the output account
 
+### Borrowing Process
+
+1. **Credit Account Check**: Verifies the existence of a credit account for the input address
+2. **Borrow Execution**: Executes the borrow action through the Mars credit account, which:
+   - Borrows the specified amount of the given denom
+   - Transfers the borrowed tokens to the output account specified in the library configuration
+3. **Token Transfer**: The borrowed tokens are automatically transferred to the output account
```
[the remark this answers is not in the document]
jonas089: done

uditvira (approved):

PR #408, opened 2025-07-02
Implementing magma protocol for providing liquidity

markoj2207:
This libraries implements Magma protocol for liquidity providing. The **Valence Magma LPer library** allows users to deposit into a Magma Vault while the **Valence Magma Withdrawer library** allows users to withdraw liquidity from Magma Vault.


contracts/libraries/magma-lper/README.md
```diff
@@ -0,0 +1,51 @@
+# Magma lper library
```
keyleu, line 1:
```suggestion
# Magma LPer Library

File: contracts/libraries/magma-lper/src/contract.rs [hunk 2]
```diff
@@ -0,0 +1,109 @@
...
+    match msg {
+        FunctionMsgs::ProvideLiquidity {
+            token_min_amount_0,
+            token_min_amount_1,
+        } => {
+            let deposit_msg = valence_magma_utils::msg::DepositMsg {
+                amount0_min: token_min_amount_0.unwrap_or_default(),
+                amount1_min: token_min_amount_1.unwrap_or_default(),
+                to: cfg.output_addr.to_string(),
+            };
```
keyleu, line 65: why are the min amounts needed? Can't we reuse the balance of the input accounts?

contracts/libraries/magma-withdrawer/README.md
```diff
@@ -0,0 +1,41 @@
+# Magma withdrawer library
```
keyleu, line 1:
```suggestion
# Magma Withdrawer Library

File: contracts/libraries/magma-withdrawer/README.md [hunk 4]
```diff
@@ -0,0 +1,41 @@
...
+
+```rust
+pub struct LibraryConfig {
+    // Address of the input account 
+    pub input_addr: LibraryAccountType,
+    // Address of the output account 
+    pub output_addr: LibraryAccountType,
+    // Address of the vault we are going to withdraw liquidity from 
+    pub vault_addr: String,
+}
```
keyleu, line 40: this config is not right, should be
```rust
pub struct LibraryConfig {
    /// Address of the input account
    pub input_addr: LibraryAccountType,
    /// Address of the output account
    pub output_addr: LibraryAccountType,
    /// Configuration for the liquidity provider
    /// This includes the pool address and asset data
    pub lp_config: LiquidityProviderConfig,
}
```

contracts/libraries/magma-withdrawer/src/msg.rs
```diff
@@ -0,0 +1,116 @@
...
+#[cw_serde]
+#[derive(ValenceLibraryInterface)]
+pub struct LibraryConfig {
+    /// Address of the input account
+    pub input_addr: LibraryAccountType,
+    /// Address of the output account
+    pub output_addr: LibraryAccountType,
+    /// Configuration for the liquidity provider
+    /// This includes the pool address and asset data
+    pub lp_config: LiquidityProviderConfig,
```
keyleu, line 36: The comment and variable are not right. Should be liquidity withdrawer and variable
withdrawer_config: LiquidityWithdrawerConfig

docs/src/libraries/cosmwasm/magma_lper.md
```diff
@@ -0,0 +1,65 @@
+# Magma lper library
```
keyleu, line 1:
```suggestion
# Magma LPer Library

File: docs/src/libraries/cosmwasm/magma_withdrawer.md [hunk 7]
```diff
@@ -0,0 +1,57 @@
+# Magma withdrawer library
```
keyleu, line 1:
```suggestion
# Magma Withdrawer Library

File: docs/src/libraries/cosmwasm/magma_withdrawer.md [hunk 8]
```diff
@@ -0,0 +1,57 @@
...
+
+```rust
+pub struct LibraryConfig {
+    // Address of the input account 
+    pub input_addr: LibraryAccountType,
+    // Address of the output account 
+    pub output_addr: LibraryAccountType,
+    // Address of the vault we are going to withdraw liquidity from 
+    pub vault_addr: String,
+}
```
keyleu, line 44: This one is wrong, probably taken from another library

contracts/libraries/magma-withdrawer/src/contract.rs
```diff
@@ -0,0 +1,119 @@
...
+    _env: Env,
+    _info: MessageInfo,
+    msg: FunctionMsgs,
+    cfg: Config,
+) -> Result<Response, LibraryError> {
+    match msg {
+        FunctionMsgs::WithdrawLiquidity {
+            token_min_amount_0,
+            token_min_amount_1,
+            shares,
```
keyleu, line 61: We don't need to pass the shares here, we can always withdraw the entire balance of the input account.
Also, regarding the token_min_amounts, do we need these? Can't we just use the balance of the input account

keyleu (changes requested):
Hey there! Thanks for the first contribution!

Left some comments regarding the inconsistencies with the API and docs and some suggestion/question with the parameters.

Also if you can please run clippy/fmt to make CI pass that would be great.

contracts/libraries/magma-withdrawer/src/contract.rs
markoj2207: The idea of having shares as an input parameter was to enable partial liquidations. If this is not needed - we can indeed withdraw the entire shares. Similarly, for min amounts, they provide possibility to specify the min amounts we expect. But, again, we can only support full withdraw with all shares we have and also hardcode min amounts to zero.

contracts/libraries/magma-lper/src/contract.rs
markoj2207: Yes, similarly to my other reply in withdrawal case, if we don't need to support ability to specify the min amounts so that tx fail if we cannot deposit at least the min amounts - we can completely omit that and actually try to deposit funds that are inside the input account

keyleu: yup we don't need to. In our model if we want to do partial liquidations we would forward the partial amount to the input account with a forwarder library. We always "consume" the full resources of the input account.

keyleu: yup we don't need to. In our model if we want to do partial liquidations we would forward the partial amount to the input account with a forwarder library. We always "consume" the full resources of the input account. So we can put the min_amounts the same as balance I guess

keyleu: yup we don't need to. In our model if we want to do partial liquidations we would forward the partial amount to the input account with a forwarder library. We always "consume" the full resources of the input account. We do the same in all our current LPers / Withdrawers - consume the full amounts.

contracts/libraries/magma-withdrawer/src/contract.rs
keyleu: Like answered above, we always consume the full amount of the input account. I don't think it's necessary to check how much we are going to get, as we are not doing that in any withdrawer so we can set these to 0.

docs/src/libraries/cosmwasm/magma_lper.md
```diff
@@ -0,0 +1,65 @@
...
+  IA -- 4/Provide Liquidity
+                  [Tokens] --> M
+  M -- 4'/Mint Shares --> OA
+
+```
+## Functions
+
+| Function    | Parameters | Description |
+|-------------|------------|-------------|
+| **ProvideLiquidity** | `amount_0_min: Option<Uint128>` <br>`amount_1_min: Option<Uint128>` | Provide double-sided liquidity to the pre-configured **Magma Vault** from the **input account**, and receive the **shares** into the **output account**. |
```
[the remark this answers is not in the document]
markoj2207: Done.

contracts/libraries/magma-lper/README.md
markoj2207: Done

contracts/libraries/magma-withdrawer/README.md
markoj2207: Done.

docs/src/libraries/cosmwasm/magma_lper.md
markoj2207: Done

docs/src/libraries/cosmwasm/magma_withdrawer.md
markoj2207: Code in implementation is updated to reflect this.

keyleu:
looks good now @markoj2207 , can make clippy pass as well to run CI before merging? ty!

keyleu:
can you merge latest main branch? Fixed clippy failing in the CI so it should pass now

keyleu (approved):
LGTM! Couldn't merge because merging the duality one caused some small conflicts

PR #410, opened 2025-07-08
fix: Magma libraries corrections

markoj2207:
Two fixes:
- changed magma deposit and withdrawal input parameters from Uint128 to String.
- execute msgs variants fix - included Deposit/Withdraw on msg creation.


keyleu:
can you regen the schemas please? @markoj2207

PR #417, opened 2025-07-14
audit: don't allow adding/removing registries with no verification gateway

keyleu:


solidity/src/authorization/Authorization.sol
bekauz, line 1: would it make sense to be doing some checks on `updateVerificationGateway` function in case that verification gateway is associated with some registries?

keyleu: don't think so?

The updateVerificationGateway is more to disable/enable ZK execution than to provide a new one.

If the admin decides to really change from one VerificationGateway to another he needs to re-add all registries that were previously added.

We're already allowing overwriting on registries so it's not a problem to do so.

Also if admin wants to keep the state clean he can also remove all the registries in an additional call so that shouldn't be an issue either.

bekauz (approved):

PR #434, opened 2025-08-06
refactor: zk authorization modifications and verification router + switch

keyleu:
refactors the ZK authorization contract and overall ZK verification flow according to the discussion we had. This applies to both CW and EVM.

In summary: 
-Adds a new String parameter (route) to the ZKAuthorization and removes the previously added tag. 
-The verification gateway is now removed and we introduce a VerificationRouter and a Verifier/Switch. The verification router is an append-only contract that contains a list of routers with their specific verifiers and the switch/verifier is an immutable contract that stores a domainVK.
- Domain inputs are not passed anymore and are now constructed using the first 32 bytes of the public inputs (therefore the comparison is not needed anymore).
- Domain proof is renamed to "payload" as we might eventually remove that if we move to the DA direction.

Has been tested with real proofs/inputs and real domain VK taken from our current coprocessor.


solidity/src/authorization/Authorization.sol
```diff
@@ -678,14 +667,14 @@ contract Authorization is Ownable, ICallback, ReentrancyGuard {
...
+        VerificationRouter _verificationRouter = verificationRouter;
+        if (address(_verificationRouter) == address(0)) {
+            revert("Verification router not set");
         }
 
-        // Verify the proof using the verification gateway
-        if (!verificationGateway.verify(decodedZKMessage.registry, _proof, _message, _domainProof, _domainMessage)) {
+        // Verify the proof using the verification router with the specific route provided in the authorization
+        if (!verificationRouter.verify(zkAuthorizationData.route, zkAuthorizationData.vk, _proof, _inputs, payload)) {
             revert("Proof verification failed");
```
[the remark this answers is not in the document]
keyleu: fixed!

solidity/src/verification/interfaces/Verifier.sol
```diff
@@ -0,0 +1,23 @@
...
+     * @dev Verifies a proof against the given inputs with a given VK.
+     * @param vk The verification key to use for the proof.
+     * @param proof The proof to verify.
+     * @param inputs The inputs to verify against.
+     * @param payload Additional payload, which in first version will be the domain proof to be validated against the domain VK.
+     * @return True if the proof is valid, false otherwise.
+     */
+    function verify(bytes calldata vk, bytes calldata proof, bytes calldata inputs, bytes calldata payload)
+        external
+        returns (bool);
```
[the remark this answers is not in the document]
keyleu: fixed!

contracts/middleware/verification/router/src/msg.rs
```diff
@@ -1,32 +1,38 @@
...
+
 use cosmwasm_schema::{cw_serde, QueryResponses};
-use cosmwasm_std::Binary;
+use cosmwasm_std::{Addr, Binary};
 use cw_ownable::{cw_ownable_execute, cw_ownable_query};
 
 #[cw_serde]
 pub struct InstantiateMsg {
-    pub domain_vk: Binary,
+    pub initial_routes: HashMap<String, String>,
```
bekauz, line 9: nice, no need to check for duplicates this way

contracts/middleware/verification/router/src/contract.rs
```diff
@@ -0,0 +1,128 @@
...
+        }
+        QueryMsg::Verify {
+            route,
+            vk,
+            inputs,
+            proof,
+            payload,
+        } => to_json_binary(&verify(deps, route, vk, inputs, proof, payload)?),
+    }
+}
```
bekauz, line 91: wonder if it would be useful to add a verifier_routes query, e.g. if something goes wrong with a particular verifier and we needed to fetch all routes that are pointing to it?

contracts/middleware/verification/router/src/state.rs
```diff
@@ -0,0 +1,5 @@
+use cosmwasm_std::Addr;
+use cw_storage_plus::Map;
+
+// Routes to different verifier contracts
+pub const ROUTES: Map<String, Addr> = Map::new("routes");
```
bekauz, line 5: maybe this can be an indexedmap if we decide we want to fetch all routes pointing to a given verifier

contracts/middleware/verification/router/src/tests.rs
bekauz, line 1: can we add one simple test that touches the verification query? just to make sure the verify query stays intact

contracts/middleware/verification/sp1-verifier/src/error.rs
```diff
@@ -0,0 +1,8 @@
+use cosmwasm_std::StdError;
+use thiserror::Error;
+
+#[derive(Error, Debug, PartialEq)]
+pub enum ContractError {
+    #[error("{0}")]
+    Std(#[from] StdError),
+}
```
bekauz, line 8: is this used?

program-manager/src/domain/cosmos_cw.rs
```diff
@@ -953,7 +953,7 @@ impl CosmosCosmwasmConnector {
         hash: String,
         retry: u64,
         err: String,
-    ) -> BoxFuture<Result<GetTxResponse, CosmosCosmwasmError>> {
+    ) -> BoxFuture<'_, Result<GetTxResponse, CosmosCosmwasmError>> {
```
bekauz, line 956: clippy?

contracts/middleware/verification/router/src/state.rs
keyleu: don't think it's worth the complication tbh, we should only have set 1 verifier for a route. There's no point have multiple routes to the same verifier

contracts/middleware/verification/router/src/tests.rs
keyleu: added!

PR #440, opened 2025-09-30
feat: erc7540 vault deposit flow

bekauz:
# Description

This is an intermediate pr for the updated two-way vault design that focuses on the deposit flow.
The follow-up pr will implement the withdraw flow as well as any leftover logic relating to owner operations, storage optimizations, proxy, etc.
Putting this out here now so that it's easier to review the core logic.

new additions/changes from the previous vaults:
- notion of `operator`: vault now implements `IERC7540Operator`, which describes the logic for users to approve other addresses (operators) to issue requests on their behalf
- pausing due to stale redemption rate/share price is now implicit: when calling deposit, `onlyWhenSharePriceNotStale` modifier will check for the last update timestamp and revert in case the price is stale (to be discussed)
  - this way if the vault stops accepting deposits due to stale price, the only thing needed to resume the normal operations will be to run the strategist and update the share price

### DIFF
--- a/solidity/src/vaults/ValenceXCV.sol
+++ b/solidity/src/vaults/ValenceXCV.sol
@@ -3,12 +3,20 @@ pragma solidity ^0.8.28;
 
 import {ERC4626Upgradeable} from "@openzeppelin-contracts-upgradeable/token/ERC20/extensions/ERC4626Upgradeable.sol";
 import {BaseAccount} from "../accounts/BaseAccount.sol";
-
+import {Math} from "@openzeppelin/contracts/utils/math/Math.sol";
 import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import {OwnableUpgradeable} from "@openzeppelin-contracts-upgradeable/access/OwnableUpgradeable.sol";
 import {Initializable} from "@openzeppelin-contracts-upgradeable/proxy/utils/Initializable.sol";
+import {SafeERC20} from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+
+// TODO: look into any ERC-4626 defaults and think if anyhthing else needs
+// to be overridden
 
 contract ValenceXCV is Initializable, ERC4626Upgradeable, OwnableUpgradeable {
+    using Math for uint256;
+
+    uint256 internal ONE_SHARE;
+
     // current share price
     uint256 public sharePrice;
 
@@ -24,6 +32,7 @@ contract ValenceXCV is Initializable, ERC4626Upgradeable, OwnableUpgradeable {
 
     error DepositAccountNotSet();
     error StrategistNotSet();
+    error ZeroDepositAmount();
 
     event SharePriceUpdated(uint256 indexed sharePrice);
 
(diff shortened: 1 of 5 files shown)


solidity/src/vaults/ValenceXCV.sol
```diff
@@ -0,0 +1,265 @@
...
+
+    /// The address of the authorized strategist who can update the share price.
+    address public strategist;
+
+    /// The Valence account contract that holds the deposited funds.
+    address public depositAccount;
+
+    /// Mapping of controllers to their approved operators.
+    mapping(address controller => mapping(address operator => bool))
+        public operators;
```
keyleu: wouldnt a simple mapping(address => address) suffice? don't see why you'd need a bool. If it's present in the map it's approved if it isn't then it's not. Would save storing an additional slot. 
The check would then be checking if address is not address(0) - the default value

solidity/src/vaults/ValenceXCV.sol
```diff
@@ -0,0 +1,265 @@
...
+    /// @dev Restricts function to cases where the share price is up to date
+    modifier onlyWhenSharePriceNotStale() {
+        if (block.timestamp - lastUpdateTimestamp > sharePriceMaxAge) {
+            revert StaleSharePrice();
+        }
+        _;
+    }
+
+    constructor() {
+        // _disableInitializers();
```
keyleu: this should not be commented so that initialization is only through the proxy.

solidity/src/vaults/ValenceXCV.sol
```diff
@@ -0,0 +1,265 @@
...
+    function initialize(
+        address _owner,
+        address strategistAddress,
+        address underlying,
+        address depositAccountAddress,
+        string memory vaultTokenName,
+        string memory vaultTokenSymbol,
+        uint256 startSharePrice,
+        uint256 maxSharePriceAge,
+        uint256 maxPriceChangeBips
```
keyleu: nit: we are using Bps everywhere instead of Bips

solidity/src/vaults/ValenceXCV.sol
```diff
@@ -0,0 +1,265 @@
...
+    /// @return A boolean indicating the success of the operation (always true)
+    function setOperator(
+        address operator,
+        bool approved
+    ) external returns (bool) {
+        // set the operator status
+        operators[msg.sender][operator] = approved;
+        emit OperatorSet(msg.sender, operator, approved);
+        return true;
+    }
```
keyleu: if you still decide to go with the mapping with a bool. If you are setting it to false you should not modify but remove the entry to get gas refund instead of paying for modifying the value. (Default value when it's not present in the mapping is false, so there's no point setting it to false, just use `delete`).

solidity/src/vaults/ValenceXCV.sol
```diff
@@ -0,0 +1,265 @@
...
+    /// Checks whether a given address is an approved operator for a controller
+    /// @param controller The address of the controller
+    /// @param operator The address of the operator to check
+    /// @return A boolean indicating whether the operator is approved
+    function isOperator(
+        address controller,
+        address operator
+    ) external view returns (bool) {
+        return operators[controller][operator];
+    }
```
keyleu, line 155: This is not used right? Also the mapping is already public so there is a getter method to get values from the mapping created automatically.

If you are using it as a helper method internally should be set as internal instead of external as we are just duplicating.

solidity/src/vaults/ValenceXCV.sol
bekauz: a controller can have multiple approved operators. not exactly sure under what circumstances that would be desirable, but just staying inline with the erc7540 spec here

solidity/src/vaults/ValenceXCV.sol
bekauz: it's just for erc7540 compliance, but i just realized i didn't add tests for it so doing that now

solidity/src/vaults/ValenceXCV.sol
bekauz: done, enabled proxy in tests as well

keyleu (approved):

PR #442, opened 2025-10-07
docs: final doc version, extend EVM flow documentation

keyleu:
Removed the security email and the under development messages. Also removed the deprecated local-ic deployment docs that we never used.

Removed also some unused files that were dangling.


docs/src/components/_overview.md
```diff
@@ -8,7 +8,6 @@ Valence currently supports CosmWasm and EVM. SVM support coming soon. The rest o
     - [Accounts](../accounts/_overview.md)
     - [Libraries and Functions](./libraries_and_functions.md)
     - [Programs and Authorizations](./programs_and_authorizations.md)
-    - [Middleware](./middleware.md)
```
bekauz, line 11: removing because it's WIP?

keyleu: yeah it's WIP and there's already a Middleware section somewhere else that has it explained better
https://docs.valence.zone/middleware/_overview.html. So it kinda was duplicated

bekauz (approved):

